PR #386, opened 2015-01-02
Add artifact publishing to oss.jfrog.org with TravisCI

sgoings:
Right now this will just publish the master version of Avian with flavors:
- windows (i386 and x86_64)
- linux (i386 and x86_64)

but I'm looking to get TravisCI to let us in on their MacOSX build slave pool (for internal backwards compatibility, we're using the Jenkins avian job to publish the Mac flavors to oss.jfrog.org).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -243,6 +243,9 @@ publishing {
 }
 
 artifactoryPublish {
+  onlyIf {
+    System.env.'TRAVIS_BRANCH' == "master"
+  }
   dependsOn assemble
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -244,7 +244,7 @@ publishing {
 
 artifactoryPublish {
   onlyIf {
-    System.env.'TRAVIS_BRANCH' == "master"
+    System.env.'TRAVIS_BRANCH' == "master" && !System.env.'TRAVIS_PULL_REQUEST'.toBoolean()
   }
   dependsOn assemble
 }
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -107,12 +107,6 @@ model {
       operatingSystem SupportedOS.valueOf(platform.toUpperCase())
       architecture "${arch}"
     }
-    if(platformArch != currentPlatformArch) {
-      create(currentPlatformArch) {
-        operatingSystem SupportedOS.CURRENT
-        architecture "${currentArch}"
-      }
-    }
```
sgoings: I got rid of this because it made the Gradle build linux i386 + x86_64 if you were only wanting to publish i386 but on x86_64 arch

build.gradle
```diff
@@ -249,6 +243,9 @@ publishing {
 }
 
 artifactoryPublish {
+  onlyIf {
+    System.env.'TRAVIS_BRANCH' == "master" && !System.env.'TRAVIS_PULL_REQUEST'.toBoolean()
```
joshuawarner32: How will this affect publishing releases (presumably not done directly from master)?

sgoings: In that case we should be publishing to bintray. See https://github.com/ReadyTalk/gradle-readytalk-js/blob/master/build.gradle

joshuawarner32:
:+1: LGTM.

PR #389, opened 2015-01-05
Add mac publishing + testing

sgoings:
- Move apt calls to ci script

Currently the mac build is failing due to JAVA_HOME not being set (it is set in the Gradle script, but not in the makefile). I'd suggest that the actual build (makefile) be set up to determine this a bit better... I won't be addressing this in this pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,8 @@ env:
 
 matrix:
   fast_finish: true
+  exclude:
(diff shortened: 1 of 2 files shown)


sgoings:
Thanks for keeping me in line, @joshuawarner32. I took your latter suggestion and applied a change to the .travis.yml file.

PR #391, opened 2015-01-10
classpath-openjdk: Fix freeZipFileEntry()

lostdj:
Those ZipFile::Entries are part of ZipFile instance itself and will fail to free(). See :1504 and :1520 in openZipFile().

PR #392, opened 2015-01-11
classpath-openjdk: Fix freeZipFileEntry()

lostdj:
Those ZipFile::Entries are part of ZipFile instance itself and will fail to free(). See :1504 and :1520 in openZipFile().

Tested it on something that failed before original fix in #391 PR, it's all good.


joshuawarner32:
Thanks!

PR #397, opened 2015-01-16
fix broken interpreter build due to out-of-order class initialization

### LINKED ISSUE #396: Interpreter: Classes derived from Throwable have uninitialized vtable
Test case:
```
public class mymain
{
	public static void main(String[] args) throws Exception
	{
		try
		{
			((Object)null).getClass();
		}
		catch(Exception e)
		{
			e.printStackTrace();
		}
	}
}
```

Now, it's good enough of a fix in this case as replacing `class_` pulled from stack with `m->class_()` here: https://github.com/ReadyTalk/avian/blob/master/src/interpret.cpp#L2052 `if(!class_()->virtualTable() && m->class_()->virtualTable()) class_ = m->class_();`. But I'm really not sure about a proper fix. Why vtable isn't initialized in the first place? `initClass(t, class_);` doesn't help either.

dicej:
When we initialize the vtables for bootstrap Java classes such as
java.lang.NullPointerException (i.e. classes which the VM has built-in
knowledge of), we assign the superclass's vtable to any class which
has no declared virtual methods of its own.  However, that vtable will
be null if we haven't initialized the superclass yet.  Therefore, we
must order this process such that no class is initialized until after
all its superclasses.

Fixes #396

### DIFF
--- a/test/NullPointer.java
+++ b/test/NullPointer.java
@@ -17,6 +17,12 @@ public class NullPointer {
 
   public static void main(String[] args) {
     try {
+      ((Object) null).getClass();
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+
+    try {
       throw_(null);
       throw new RuntimeException();
     } catch (NullPointerException e) {
--- a/src/tools/type-generator/main.cpp
+++ b/src/tools/type-generator/main.cpp
@@ -1408,8 +1408,20 @@ void writeInitializations(Output* out, Module& module)
   }
 }
 
-void writeJavaInitialization(Output* out, Class* cl)
+void writeJavaInitialization(Output* out,
+                             Class* cl,
+                             std::set<Class*>& alreadyInited)
 {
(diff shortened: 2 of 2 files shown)


lostdj:
Thanks!

test/NullPointer.java
```diff
@@ -17,6 +17,12 @@ private static void throwAndCatch(Object o) {
 
   public static void main(String[] args) {
     try {
+      ((Object) null).getClass();
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+
+    try {
```
lostdj: Wait, why #396 wasn't caught here before? It looks exactly like a trap presented in the new test case.

dicej: To be honest, I don't know.  There must be some other mechanism by which a class can get a valid vtable besides the VM boot process, but I don't know what/where it is offhand.

lostdj: :>

PR #405, opened 2015-02-05
fix ARM64 iOS JNI crashes

dicej:
As documented at
https://developer.apple.com/library/ios/documentation/Xcode/Conceptual/iPhoneOSABIReference/Articles/ARM64FunctionCallingConventions.html,
the ARM64 iOS ABI differs from the generic ABI in a few important
ways.  Specifically, arguments passed via the stack are aligned
according to their natural alignment instead of 8 bytes.  The VM's
dynamic call code was aligning each argument to 8 bytes, so native JNI
code couldn't find them in their expected places.

Also, we weren't setting the "os.arch" system property on ARM64, so I
fixed that too.


dicej:
Oops, hold on.  I just noticed a bug.

dicej:
Should be good now.

PR #406, opened 2015-02-06
I added 2 missing method to java.lang.Character.

marcinolawski:
I added 2 missing method to java.lang.Character, isJavaIdentifierStart(char ch) and isJavaIdentifierPart(char ch).


joshuawarner32:
Thanks!

Couple small points:
* The `int codePoint` overloads should go along with these (similarly to many other methods in that class)
* I'd prefer seeing something much more compact like `return isLetter(ch) || ch == '$' || ch == '_';`

marcinolawski:
Done.

PR #407, opened 2015-02-06
load bootstrap classes in findInterfaceMethod

### LINKED ISSUE #404: fix lazy boostrap class initialization
In afbd4ff30303f6f205895d86e752f493b4a91e3c, I made a low-risk, but very specific fix for a more general problem: "bootstrap" classes (i.e. classes which the VM has built-in knowledge of) need to be loaded from the classpath before any of their methods are called.  Based on recent testing, I see there are more cases than I previously thought where we try to call methods on "unloaded" bootstrap classes, so we need a more general solution to the problem, preferably one that does so lazily so we don't have to load the class just because an instance is created (but no method is called)

dicej:
In afbd4ff, I made a low-risk, but very specific fix for a more
general problem: "bootstrap" classes (i.e. classes which the VM has
built-in knowledge of) need to be loaded from the classpath before any
of their methods are called.  Based on recent testing, I found there were
more cases than I previously thought where the VM tries to call methods on
"unloaded" bootstrap classes, so we needed a more general solution to
the problem.

This commit addresses it by closing the last (known) loophole by which
methods might be called on bootstrap classes: invokeinterface, and its
helper method findInterfaceMethod.  The fix is to check for bootstrap
classes in findInterfaceMethod and load the full versions if
necessary.  This process may lead to garbage collection and/or thrown
exceptions, which made me nervous about cases of direct or indirect
calls to findInterfaceMethod not expecting those events, which is why
I hadn't used that approach earlier.  However, it turns out there were
only a few places that made non-GC-safe calls to findInterfaceMethod,
and a bit of code rearrangement fixed that.

Fixes #404

### DIFF
--- a/src/avian/machine.h
+++ b/src/avian/machine.h
@@ -2675,7 +2675,12 @@ inline GcMethod* findInterfaceMethod(Thread* t,
                                      GcMethod* method,
                                      GcClass* class_)
 {
-  assertT(t, (class_->vmFlags() & BootstrapFlag) == 0);
+  if (UNLIKELY(class_->vmFlags() & BootstrapFlag)) {
+    PROTECT(t, method);
+    PROTECT(t, class_);
+
+    resolveSystemClass(t, roots(t)->bootLoader(), class_->name());
+  }
 
   GcClass* interface = method->class_();
   GcArray* itable = cast<GcArray>(t, class_->interfaceTable());
--- a/src/classpath-openjdk.cpp
+++ b/src/classpath-openjdk.cpp
@@ -2452,18 +2452,9 @@ object makeJconstructor(Thread* t,
 }
 #endif  // HAVE_JexecutableHasRealParameterData
 
-void resolveBootstrap(Thread* t, Gc::Type type)
-{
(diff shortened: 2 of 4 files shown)


jentfoo:
Thanks for doing this so quickly @dicej...It looks good to me, but I am too nervous to merge it because I am not as familiar in the C portions.  @joshuawarner32, do you want to give it a second opinion?

joshuawarner32:
I'm a little uncertain how this works for the `findVirtualMethod` case.  (@dicej?)

However, I think these are pretty safe changes, and I'm going to assume that test failed before the corresponding vm changes.  So I'm find with merging.

dicej:
Yeah, I was wondering about `findVirtualMethod`, too, and wrote a test for it, but couldn't make it fail.  Then I looked for where the class was being loaded in GDB, and found that `compileVirtualMethod2` loads the class at the last moment, which reminded me that bootstrap classes *do* have vtables before they are loaded, just not interface vtables (see `bootJavaClass` for details).  So we don't need to do anything special in `findVirtualMethod`.

PR #409, opened 2015-02-08
Added missing methods to ByteArrayOutputStream, Modifier and UTFDataFormatException.

marcinolawski:
I added: 
- ByteArrayOutputStream.writeTo(OutputStream)
- UTFDataFormatException
- Modifier.isTransient(int)

### DIFF
--- a/classpath/java/io/ByteArrayOutputStream.java
+++ b/classpath/java/io/ByteArrayOutputStream.java
@@ -13,7 +13,8 @@ package java.io;
 public class ByteArrayOutputStream extends OutputStream {
   private static final int BufferSize = 32;
 
-  private Cell chain;
+  private Cell firstCell;
+  private Cell curCell;
   private int length;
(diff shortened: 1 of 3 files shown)


joshuawarner32:
Nit: the indentation is a weird combination of spaces and tabs.  Please use 2 spaces to indent.

Also, it's not a huge problem, but I'm not a big fan of copying the `Cell`s into a temporary array.  If it's not too much trouble, could you swap around the directionality of the links in `Cell` (probably using `firstCell` and `curCell` fields), so we can iterate in the natural order, and not have to build that temporary copy?  If you're in a time crunch, don't worry about it.

I was also going to suggest using recursion to cleverly reverse it on the stack, but I have no evidence that would be actually faster in practice.

marcinolawski:
Done.

joshuawarner32:
Thanks!

PR #413, opened 2015-02-19
Fix delete windows

brandonecraig:
Fixing java.io.File.delete() for windows.

Adding gradle so the build system is comparable to master.


joshuawarner32:
@brandonecraig - can you make a PR to master with _just_ the change in question?  We can let travis verify it there, and then we can tackle backporting it and setting up a release branch, as necessary.

Nit: I really don't like the preprocessor stuff and C++ expressions breaking out of a strict tree structure; I think it hurts readability.  I'd much rather see the `REMOVE` like duplicated.

Another, smaller nit:  I'd prefer using GetFileAttributes instead of PathIsDirectory, so we don't have to link in extra windows libs.

PR #414, opened 2015-02-19
fix java.io.File.delete() implementation for windows

brandonecraig:
On windows, check if  the path is a directory. If so use RemoveDirectory.


joshuawarner32:
Thanks, @brandonecraig!

PR #425, opened 2015-03-16
use size_t instead of unsigned in a bunch of appropriate places

joshuawarner32:
This would theoretically break compatibility with apps using embedded
classpaths, on big-endian architectures - because of the size type
extension.  However, we don't currently support any big-endian
architectures, so it shouldn't be a problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,7 +454,7 @@ setting the boot classpath to "[bootJar]".
       extern const uint8_t SYMBOL(end)[];
 
       EXPORT const uint8_t*
-      bootJar(unsigned* size)
+      bootJar(size_t* size)
       {
         *size = SYMBOL(end) - SYMBOL(start);
         return SYMBOL(start);
@@ -682,7 +682,7 @@ containing them.  See the previous example for instructions.
       extern const uint8_t BOOTIMAGE_BIN(end)[];
 
       EXPORT const uint8_t*
-      bootimageBin(unsigned* size)
+      bootimageBin(size_t* size)
       {
         *size = BOOTIMAGE_BIN(end) - BOOTIMAGE_BIN(start);
         return BOOTIMAGE_BIN(start);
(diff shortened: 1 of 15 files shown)


dicej:
https://travis-ci.org/ReadyTalk/avian/jobs/54638514#L953

dicej:
Once again, Travis tells us the build is fine when it's not.  Boo.  Hiss.

joshuawarner32:
Good thing you look at the output; I haven't been.  When running ./test/ci.sh locally, it does what we want (fails on errors), so it's either travis's fault, or it's something weird we're doing in our configuration (.travis.yml).  @sgoings, could you take a look?

dicej:
Actually, I'm able to reproduce this with test/ci.sh locally, too.  It looks like it might be related to our use of && and || to do conditionals instead of if/then, with the result that when "make mode=debug bootimage=true test" fails, it doesn't error out of the script; it only skips the next item in the && chain.

dicej:
http://stackoverflow.com/questions/25794905/why-does-set-e-true-false-true-not-exit

dicej:
This seems to fix it:

```
diff --git a/test/ci.sh b/test/ci.sh
index 4c99e1d..69c89d2 100755
--- a/test/ci.sh
+++ b/test/ci.sh
@@ -86,17 +86,22 @@ else
 
   make_target=test
 
-  (! has_flag arch) && run make ${flags} jdk-test
+  if ! has_flag arch; then
+    run make ${flags} jdk-test
+  fi
+
   run make ${flags} ${make_target}
   run make ${flags} mode=debug ${make_target}
   run make ${flags} process=interpret ${make_target}
 
-  (has_flag openjdk-src || ! has_flag openjdk) && \
-    run make ${flags} mode=debug bootimage=true ${make_target} && \
+  if has_flag openjdk-src || ! has_flag openjdk; then
+    run make ${flags} mode=debug bootimage=true ${make_target}
     run make ${flags} bootimage=true ${make_target}
+  fi
 
-  (! has_flag openjdk && ! has_flag android && ! has_flag arch) && \
+  if ! has_flag openjdk && ! has_flag android && ! has_flag arch; then
     run make ${flags} openjdk=$JAVA_HOME ${make_target}
+  fi
 
   run make ${flags} tails=true continuations=true heapdump=true ${make_target}
   run make ${flags} codegen-targets=all
```

I'll make a pull request.

PR #426, opened 2015-03-16
ensure ci.sh exits with error when any test fails

### LINKED ISSUE #416: ci.sh test failures don't cause Travis job to fail
I was just skimming over the output from a recent Travis job and noticed that one of the tests failed without causing the overall job to fail. That's not cool.

Search for "fail" in https://travis-ci.org/ReadyTalk/avian/jobs/52475399 to see what I mean.

dicej:
As explained in
http://stackoverflow.com/questions/25794905/why-does-set-e-true-false-true-not-exit,
"set -e" will only cause the script to exit if the failing command is
"unhandled".  The fancy || and && shortcuts we were using to run some
of the tests caused bash to think we were handling the errors, which
isn't what we wanted.

Fixes #416

### DIFF
--- a/test/ci.sh
+++ b/test/ci.sh
@@ -86,17 +86,22 @@ else
 
   make_target=test
 
-  (! has_flag arch) && run make ${flags} jdk-test
+  if ! has_flag arch; then
+    run make ${flags} jdk-test
+  fi
+
   run make ${flags} ${make_target}
   run make ${flags} mode=debug ${make_target}
   run make ${flags} process=interpret ${make_target}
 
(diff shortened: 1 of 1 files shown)


joshuawarner32:
I'd hazard a guess to say that the first and third hunks aren't strictly necessary.  However, do show a good example, and having that precedent in there ought to discourage further similar mistakes.

FWIW, I wasn't seeing this locally because I wasn't actually making ci.sh fail how it was in #425.

Good find!

joshuawarner32:
Checking the build output this time...  And looks good!

Issue #420, opened 2015-03-12
printf for Avian classpath


bgould:
Hi guys

I've been experimenting with Avian for the past few months and I've been quite pleased with the high quality of your work.

One thing I've found as a slight impediment is the inability to use some libraries unmodified because they make use of Java's printf/java.util.Formatter/etc.  This is generally pretty easy to work around but is annoying to maintain a fork of a library for such a small thing.

I've done a clean-room pure Java clone of java.util.Formatter... it is not 100% compatible but it will at least accept any of the format patterns specified in OpenJDK's java.util.Formatter.  I purposely have not looked at the source of any other implementations of this class so I am not sure how it stacks up performance-wise but I tried to minimize object allocation/memory usage as much as possible.  I would be glad to submit it as a pull request if you guys want to incorporate it into Avian classpath (I don't think I put it on Github yet but I will if you're interested), and I can add your preferred license to it if you do.

Questions I have though:

1) Is this something you guys might to include? :)
2) If so, what level of completeness would you require?  Certain formats just don't work yet... %s, %d, %x, width, zerofill, etc are working but some of the more esoteric ones are not.  The parser will gladly accept all valid patterns, it just doesn't output the formats properly.
3) Do you have a Contributor License Agreement or anything like that that I would need to sign?

The reason it is not on Github yet is that I have a bit of code cleanup to do before I publish it, and also the license I choose is dependent on what your preference would be.  The guts of the implementation is in 1 source file outside of the java.* namespaces, and I have an implementation of java.util.Formatter plus the format() and printf() methods to add to java.lang.String and java.io.PrintStream in order to integrate into Avian classpath.

Please let me know if you are interested and I will proceed accordingly.  Thanks for all your great work and your contributions to free software!

dicej:
Hi BCG,

On Thu, 12 Mar 2015, BCG wrote:

> 1) Is this something you guys might to include? :)

Yes!

> 2) If so, what level of completeness would you require?  Certain formats 
> just don't work yet... %s, %d, %x, width, zerofill, etc are working but 
> some of the more esoteric ones are not.  The parser will gladly accept 
> all valid patterns, it just doesn't output the formats properly.

Whatever you have is fine.  Avian's class library will probably always be 
incomplete in various ways, and our philosophy has been to implement only 
the minimum necessary to get a given application working.  No point in 
doing extra work until you know you need it.

> 3) Do you have a Contributor License Agreement or anything like that 
> that I would need to sign?

Nope.

[...]

PR #427, opened 2015-03-17
Added java.util.Formatter implementation. Basic/common formats work,

bgould:
Adds printf/format functionality to Avian classpath, described in Avian issue #420

### DIFF
--- a/classpath/avian/FormatString.java
+++ b/classpath/avian/FormatString.java
@@ -1,0 +1,992 @@
+/* Copyright (c) 2008-2015, Avian Contributors
+
+   Permission to use, copy, modify, and/or distribute this software
+   for any purpose with or without fee is hereby granted, provided
+   that the above copyright notice and this permission notice appear
+   in all copies.
+
+   There is NO WARRANTY for this software.  See license.txt for
+   details. */
+
+package avian;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.IllegalFormatException;
+import java.util.List;
+
+// ------------------------------------------------------------------------- //
+// things that must be done in order to call this semi-complete: 
+// ------------------------------------------------------------------------- //
+//  * get the date formatter working for individual fields at a minimum 
+// ------------------------------------------------------------------------- //
+
+/**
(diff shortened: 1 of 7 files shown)


dicej:
@bgould, it looks like the Travis build failed because the FormatStrings test uses the avian.FormatString class directly, meaning the test won't build or run against any class library except the Avian one.  Since the Avian VM is also meant to work with the OpenJDK and Android class libraries, we'd prefer that all the tests also build and run against those libraries.  Is there a way the test could be rewritten to use java.util.Formatter instead of using FormatString directly?

bgould:
For sure, no problem.  Didn't realize that, sorry.

I'm not so familiar with Github... do I need to create a new pull request after I update or is there a way to attach a new commit to this one?

dicej:
Just push the commit to your master branch, and it will automatically be included in the pull request.

classpath/avian/FormatString.java
```diff
@@ -0,0 +1,992 @@
...
+     
+      default:
+        throw new IllegalFormatException(
+          "Invalid character encountered while parsing specifier: " + c);
+      }
+    }
+
+    @Override
+    public String toString() {
+      StringBuilder sb = new StringBuilder();
```
dicej: minor nit: this would be more concise and just as efficient using string concatenation (e.g. return "{ source = '" + _source + "', conversion = " + _conversion + ", flags = " ...)

joshuawarner32: To expand on that, in case you weren't aware, javac compiles string concatenations into StringBuilder .append invocations, all on the same instance (exactly what you're doing here).

bgould: That sounds good, old habits die hard I guess. Its one of those cargo-cult habits that I haven't been able to shake. I just pushed the change into my fork, I guess I might have to open another pull request

dicej: Yeah, no big deal.  I went ahead and merged it as-is, since I didn't notice any serious problems.  Feel free to open another pull request if you'd like.

PR #432, opened 2015-05-03
Fix openjdk-src build after #425

joshuawarner32:


makefile
```diff
@@ -1295,7 +1295,6 @@ ifeq ($(process),compile)
 
 	vm-asm-sources += $(src)/compile-$(arch).$(asm-format)
 endif
-cflags += -DAVIAN_PROCESS_$(process)
```
joshuawarner32: Unrelated, but I happened to notice this wasn't used anywhere.

PR #433, opened 2015-05-04
fix unsigned->size_t build regression

dicej:


src/tools/bootimage-generator/main.cpp
```diff
@@ -1861,7 +1861,8 @@ void writeBootImage2(Thread* t,
                              bootimageData.length,
                              &bootimageLength);
 
-      fprintf(stderr, "compressed heap size %d\n", bootimageLength);
+      fprintf(stderr, "compressed heap size %d\n",
+              static_cast<int>(bootimageLength));
```
joshuawarner32: FYI, the `%zu` formatter is designed specifically for size_t values.

PR #434, opened 2015-05-07
fix concurrency bugs on ios+arm64 in `enter`

joshuawarner32:
At first, it might look like the atomicIncrement operations here,
since they resolve to OSAtomicCompareAndSwap32Barrier, ought to
provide all the memory barrier guarantees we need; however, it turns
out it's not quite sufficient.

Here's why: Apple's OSAtomic\*Barrier operations guarantee that memory
operations *before* the atomic op won't be reordered with memory
operations *after* the atomic op - but makes no guarantee that the
atomic op itself won't be reordered with other operations before or
after it.  The key is that the atomic operation is not really atomic,
but rather consists of separate read, check and write steps - in a
loop, no less.  Here, presumably, the read of t->m->exclusive is
hoisted by the out-of-order processor to between the read and write
steps of the "atomic" write.

Longer term, it probably makes sense to replace this with the c11
<stdatomic.h> operations or the c++11 <atomic> types.  We ought to
actually use the atomic increment operations provided there.  As it
is, our atomicIncrement looks substantially less efficient than those,
since it's actually implemented on arm64 as two nested loops (one in
atomicIncrement and one in the compare-and-swap) instead of one.  We
should also evaluate whether the various instances of atomic
operations actually need as strong of barriers as we're giving them.

FWIW, the gcc __sync_* builtins seem to have the same problem, at
least on arm64.

### DIFF
--- a/src/machine.cpp
+++ b/src/machine.cpp
@@ -4040,6 +4040,8 @@ void enter(Thread* t, Thread::State s)
 
       t->state = s;
 
+      STORE_LOAD_MEMORY_BARRIER;
+
       if (t->m->exclusive) {
         ACQUIRE_LOCK;
 
@@ -4090,6 +4092,8 @@ void enter(Thread* t, Thread::State s)
       INCREMENT(&(t->m->activeCount), 1);
 
       t->state = s;
+
+      STORE_LOAD_MEMORY_BARRIER;
 
       if (t->m->exclusive) {
         // another thread has entered the exclusive state, so we


dicej:
Should we consider just adding calls to ```storeLoadMemoryBarrier``` in ```atomicCompareAndSwap*```?  Or have you already looked at other uses of ```atomicCompareAndSwap*``` and determined that they don't expect an implicit memory barrier?

joshuawarner32:
From the reading I did, it sounds like `storeLoadMemoryBarrier` the most expensive one, by a substantial margin - so I don't think we should just throw it around.

Other places we use `atomicCompareAndSwap*`:
* monitor\* functions in machine.h - I'm going to assume, since there's a reference to a formal algorithm, and that they're already littered with memory barriers - that they're already taken care of.
* uses of `setFlag` and `clearFlag` in machine.h - of those, the cases involving `Thread::WaitingFlag` (2900, 2931, 3008) look slightly suspicious, since they're all dealing with `t->waitNext` around there too, but I don't think there's anything funny actually going on.  I can't be sure, though.
* Implementations of `sun.misc.Unsafe` in builtin.cpp - I can't find documentation on what the memory barrier / ordering requirements on those functions should be.  Openjdk may be the definitive reference there.  However, in practice, any instructions that might be hoisted wrongly into the middle of those atomic operations will be sufficiently far removed that I don't think any current CPUs will take advantage of it. For reference, Haswell has a 192 micro-op Reorder Buffer.  Arm chips will likely have much shallower pipelines, and keep in mind, instructions will often expand into multiple micro-ops.
* Avian's `Atomic.compareAndSwapObject` implementation in classpath-avian.cpp.  This is similar.
* `markAtomic` in heap.cpp - I think this all looks fine.

Basically, I'd rather place barriers where they're called for rather than use the heavy-handed approach.  But it's definitely an open question whether there are other places we need barriers.

dicej:
Yeah, I guess I never consciously wrote code that assumed atomic operations were also barriers as far as the atomic write itself was concerned, so we *should* be okay.  If we see further concurrency issues on ARM64, we'll know where to start looking, though.

PR #435, opened 2015-06-02
Some small tunes for avian-pack

bigfatbrowncat:
I'm going to release avian-pack (a stable version with Avian 1.2.0 included). So I want to synchronize my build script with your.

Changes:
1. I had to remove -Werror everywhere. Unfortunately, some of checks in the code generate warnings on OS X (for example, clang doesn't like check "this == 0"), that leaded to a build failure.

2. I changed javac options that it could compile all the files well (I had some problems without that). This doesn't change the result, just a build tool tuning.

3. I added special options to link statically against pthread on Windows cause avian.exe shouldn't have any external dependencies. As I understand, MinGW uses non-GPL pthread implementation. Actually, improved version of this one: http://locklessinc.com/downloads/winpthreads.h, so no legal issues could apply. If I'm not right, please, explain, why.

P.S. As soon as I support the "official" Avian + Android classpath version, would you like to put a download link for my binaries (which I'll make right after you accept this pull request) on your site?

Thank you anyway.


joshuawarner32:
A few things:
* Are you up to date with this repository?  I fixed a few errors that occurred with the new clang that ships with xcode.  I have a fully up-to-date xcode and the build works fine for me.
* I'd rather not remove `-Werror` - but if that's absolutely necessary, I'd like to at least preserve it on travis builds.
* If you look at the travis failures, the first one (https://travis-ci.org/ReadyTalk/avian/jobs/65093224) is from a known issue, that almost certainly unrelated to your changes.  It might go away if the build is re-run.  The second one bears looking at (https://travis-ci.org/ReadyTalk/avian/jobs/65093224).

makefile
```diff
@@ -1748,7 +1748,7 @@ $(build)/android.dep: $(luni-javas) $(dalvik-javas) $(libart-javas) \
 		| (cd $(build)/android-src && $(jar) x)
 #	(cd android && $(jar) c *)	| (cd $(build)/android-src && $(jar) x)
 	find $(build)/android-src -name '*.java' > $(build)/android.txt
-	$(javac) -Xmaxerrs 1000 -d $(build)/android @$(build)/android.txt
+	$(javac) -J-Xms256m -J-Xmx256m -Xmaxerrs 1000 -d $(build)/android @$(build)/android.txt
```
joshuawarner32: What behavior were you seeing that you needed these?  Setting a fixed, relatively small, limit seems like asking for trouble in the future.

makefile
```diff
@@ -863,7 +863,7 @@ ifeq ($(platform),windows)
 	exe-suffix = .exe
 	rpath =
 
-	lflags = -L$(lib) $(common-lflags) -lws2_32 -liphlpapi -mconsole
+	lflags = -L$(lib) -static-libgcc -static-libstdc++ -Wl,-Bstatic -lstdc++ -lpthread -Wl,-Bdynamic $(common-lflags) -lws2_32 -liphlpapi -mconsole
```
joshuawarner32: What requires libstdc++ in the build?  Once upon a time, at least, we didn't depend on that.  I'd like to keep it that way, if possible.

joshuawarner32:
One more thing: it looks like you're unconditionally linking pthreads on windows.  IIRC, that should only be necessary for the android classpath (not avian or openjdk classpaths).  Can you make that option specific to building avian+android+windows?

Lastly, once we figure out the rest of the issues, could you force-push a new version of these changes, squashed into a single commit (or small handful, as appropriate)?

makefile
bigfatbrowncat: I haven't set a "relatively small" limit. I made it bigger than it was set by default cause it couldn't compile all the android classes without this option.
Look:
    java -XX:+UnlockExperimentalVMOptions -XX:+UnlockDiagnosticVMOptions -XX:+PrintFlagsFinal 2>&1 | grep MaxPermSize
    uintx MaxPermSize                               = 85983232        {pd product}

makefile
bigfatbrowncat: I put these options to a wrong line. I should move them to "android cp + windows" case

bigfatbrowncat:
Agreed about pthread.

About squashing - that would not be easy cause I did some merges in my repo...

joshuawarner32:
@bigfatbrowncat - worst case, squashing can be made really easy with diff/apply, like this:

```
git checkout upstream/master
git pull
git checkout -b avian-pack-a
git merge avian-pack
git checkout upstream/master
git checkout -b avian-pack-b
git diff avian-pack-a..upstream/master | git apply
git push -f origin avian-pack-b:avian-pack # Or however else you want to mange your branches.
```

makefile
joshuawarner32: A couple things:
* You're looking at a different value there than you're setting:  You're reading the Perm-gen size (which is just for stuff that will never be garbage collected, and should be small) - and you're setting the max total heap size (which doesn't have a fixed default; it'll default to 1/4 of your system memory).  What you want to grep for is 'MaxHeapSize', not 'MaxPermSize'.  On my workstation, I confirmed that it reports the default max size as ~8GB (I have 32GB memory).  256MB is substantially smaller than that.

Are you perhaps compiling on a VM with very little memory (512MB, or even less)?

If just bumping your system memory isn't an option, perhaps add a `$(JAVAC_OPTS)` instead of these switches, which will by default be empty, but on your system you can add those particular arguments you need by setting that environment variable (or make parameter, or whatever).

bigfatbrowncat: Looks like it's 256MB here... Strange. I'll reinvestigate that.

    uintx MaxHeapSize                              := 257949696       {product}

joshuawarner32:
Looks good to me.  Can you squash the commits?

bigfatbrowncat:
Closing this request and opening a new one.

Issue #436, opened 2015-06-02
Wrong JNI arg on arm64-ios


tarotanaka0:
The last JNI argument is wrong in some cases on arm64-ios.
It probably depends on JNI method signature.

```java
package test;

public class Hello {
  public static void main(String[] args) {
    hello1(null, null, 1, null, 2);
    hello2(null, null, 1, null, 2, 3);
    hello3(null, null, 1, null, 2, 3, 4);
    hello4(null, null, 1, null, 2, 3, 4, 5);
  }

  private static native void hello1(Object a, Object b, int c, Object d, int e);
  private static native void hello2(Object a, Object b, int c, Object d, int e, int f);
  private static native void hello3(Object a, Object b, int c, Object d, int e, int f, int g);
  private static native void hello4(Object a, Object b, int c, Object d, int e, int f, int g, int h);
}
```

```c
JNIEXPORT void JNICALL Java_test_Hello_hello1
(JNIEnv *, jclass, jobject a, jobject b, jint c, jobject d, jint e) {
    printf("hello1 : %d %d\n", c, e);
}

JNIEXPORT void JNICALL Java_test_Hello_hello2
(JNIEnv *, jclass, jobject a, jobject b, jint c, jobject d, jint e, jint f) {
    printf("hello2 : %d %d %d\n", c, e, f);
}

JNIEXPORT void JNICALL Java_test_Hello_hello3
(JNIEnv *, jclass, jobject a, jobject b, jint c, jobject d, jint e, jint f, jint g) {
    printf("hello3 : %d %d %d %d\n", c, e, f, g);
}

JNIEXPORT void JNICALL Java_test_Hello_hello4
(JNIEnv *, jclass, jobject a, jobject b, jint c, jobject d, jint e, jint f, jint g, jint h) {
    printf("hello4 : %d %d %d %d %d\n", c, e, f, g, h);
}
```

Output:
```
hello1 : 1 2
hello2 : 1 2 3
hello3 : 1 2 3 1875909232
hello4 : 1 2 3 4 5
```

tarotanaka0:
Possible fix:
```
--- a/src/avian/arm.h
+++ b/src/avian/arm.h
@@ -388,11 +388,11 @@ inline uint64_t dynamicCall(void* function,
     vfpIndex = VfpCount;
   }
 
-  unsigned stackSize = pad(stackIndex * BytesPerWord, 16);
[...]

PR #438, opened 2015-06-04
fix JNI stack alignment (for #436)

tarotanaka0:
This will fix corrupt JNI arg on arm64.
Tested with ios-arm64.

### DIFF
--- a/test/jni.cpp
+++ b/test/jni.cpp
@@ -101,35 +101,35 @@ extern "C" JNIEXPORT jdouble JNICALL Java_JNI_addMix(JNIEnv*,
 }
 
 extern "C" JNIEXPORT jint JNICALL
-    Java_JNI_addStackBoundary2(JNIEnv* e, jclass c, jobject o1, jobject o2, jobject o3,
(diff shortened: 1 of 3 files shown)


test/jni.cpp
```diff
@@ -100,6 +100,41 @@ extern "C" JNIEXPORT jdouble JNICALL Java_JNI_addMix(JNIEnv*,
          + a14 + a15 + a16 + a17 + a18 + a19 + a20;
 }
 
+extern "C" JNIEXPORT jint JNICALL
+    Java_JNI_addStackBoundary2(JNIEnv* e, jclass c, jobject o1, jobject o2, jobject o3,
```
joshuawarner32: Travis is giving errors on these because we have `-Wunused-parameters` and `-Werror` turned on.  I'm not a huge fan of that warning, but the fix here is simple: either put `UNUSED` after each parameter name (e.g. `JNIEnv* e UNUSED`) or just delete the name (just `JNIEnv*`).

tarotanaka0:
Fixed :)

joshuawarner32:
Sorry for the delay.  It took a little bit to set get my arm64 qemu vm going again.

Thanks, @tarotanaka0!

Issue #429, opened 2015-04-08
Trace test fails occasionally and unpredictably


dicej:
This causes ci.sh to fail randomly for unrelated pull requests, maybe 1% of the time.  One way to try reproducing it is with e.g.

```
while make mode=debug bootimage=true input=Trace run >/tmp/log.txt 2>&1; do echo success; done
```

If that loop exits, it means it fails, and hopefully /tmp/log.txt has details.  You might also want to run ```ulimit -c unlimited``` to ensure the core is dumped if it crashes.

dicej:
I've had the above loop running for the past few hours and have yet to hit a failure.  Maybe there's something about Travis's build environment that makes it more likely.

PR #441, opened 2015-06-23
fixed abs() compilation for Android classpath on OSX

bigfatbrowncat:
Trying to build Android class path (avian-pack) on OS X. Having a problem:
```
src/classpath-android.cpp:1209:10: error: taking the absolute value of unsigned type 'uintptr_t' (aka 'unsigned long') has no effect [-Werror,-Wabsolute-value]
  return llabs(arguments[0]);
         ^
src/classpath-android.cpp:1209:10: note: remove the call to 'llabs' since unsigned values cannot be negative
  return llabs(arguments[0]);
         ^~~~~
src/classpath-android.cpp:1215:22: error: using integer absolute value function 'abs' when argument is of floating point type [-Werror,-Wabsolute-value]
  return floatToBits(abs(bitsToFloat(arguments[0])));
                     ^
src/classpath-android.cpp:1215:22: note: use function 'std::abs' instead
  return floatToBits(abs(bitsToFloat(arguments[0])));
                     ^~~
                     std::abs
src/classpath-android.cpp:1215:22: note: include the header <cmath> or explicitly provide a declaration for 'std::abs'
2 errors generated.
```
This is the solution. Although we are not linking against standard library, we are using <code>abs</code> and <code>llabs</code> from it. Additionally i've fixed a nasty type conversion bug.

### DIFF
--- a/src/classpath-android.cpp
+++ b/src/classpath-android.cpp
@@ -8,6 +8,9 @@
    There is NO WARRANTY for this software.  See license.txt for
    details. */
 
+#include <cmath>
+using namespace std;
+
 struct JavaVM;
 struct _JavaVM;
 struct _JNIEnv;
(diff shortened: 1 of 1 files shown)


bigfatbrowncat:
That's strange... "Trace" test failed, but I changed nothing except <code>classpath-android.cpp</code>, so that shouldn't be my fault...

dicej:
@bigfatbrowncat it's probably not you're fault.  See issue #429

bigfatbrowncat:
Fixed code just as you proposed. Thanks!

PR #444, opened 2015-07-07
fix GC crash for bootimage builds

### LINKED ISSUE #443: GC crashes after Class#getResourceAsStream()
GC crashes after calling Class#getResourceAsStream()
on ios-arm-bootimage-android build.
(Does not happen with avian classpath)

Test code: (modifying hello-ios)
```
diff --git a/src/Hello.java b/src/Hello.java
index 71fb67c..b22d15f 100644
--- a/src/Hello.java
+++ b/src/Hello.java
@@ -9,6 +9,10 @@ public class Hello {
     this.peer = peer;
 
     InputStream in = getClass().getResourceAsStream("/hello.properties");
+      for (int j = 0; j < 1000; j++) {
+          System.err.println(j);
+          byte[] b = new byte[1024 * 1024];
+      }
     try {
       Properties props = new Properties();
```

Backtrace (when j==18):
```
* thread #1: tid = 0xdb258, 0x001cd544 hello`(anonymous namespace)::HeapClient::isFixed(void*) [inlined] vm::objectFixed(vm::Thread*, vm::GcObject*) at machine.h:1952, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x79e6058)
  * frame #0: 0x001cd544 hello`(anonymous namespace)::HeapClient::isFixed(void*) [inlined] vm::objectFixed(vm::Thread*, vm::GcObject*) at machine.h:1952
    frame #1: 0x001cd544 hello`(anonymous namespace)::HeapClient::isFixed(this=0x14d4a360, p=0x079e6058) at machine.cpp:3368
    frame #2: 0x00100a18 hello`(anonymous namespace)::local::update((anonymous namespace)::local::Context*, void**, void*, unsigned int, bool*) + 16 at heap.cpp:1132
    frame #3: 0x00100a08 hello`(anonymous namespace)::local::update((anonymous namespace)::local::Context*, void**, void*, unsigned int, bool*) [inlined] (anonymous namespace)::local::update2(c=0x14d4a864)::local::Context*, void*, bool*) + 38 at heap.cpp:1164
    frame #4: 0x001009e2 hello`(anonymous namespace)::local::update(c=0x14d4a864, p=0x01f94460, target=0x01f94458, offset=2, needsVisit=0x02f922fc)::local::Context*, void**, void*, unsigned int, bool*) + 34 at heap.cpp:1257
    frame #5: 0x001008ac hello`(anonymous namespace)::local::collect(c=0x14d4a864, p=0x01f94460, target=<unavailable>, offset=<unavailable>)::local::Context*, void**, void*, unsigned int) + 44 at heap.cpp:1381
    frame #6: 0x000ff536 hello`(anonymous namespace)::local::MyHeap::collect(vm::Heap::CollectionType, unsigned int, int) [inlined] (anonymous namespace)::local::collect(c=<unavailable>)::local::Context*, void*, unsigned int) + 906 at heap.cpp:1578
    frame #7: 0x000ff52a hello`(anonymous namespace)::local::MyHeap::collect(vm::Heap::CollectionType, unsigned int, int) + 134 at heap.cpp:1618
    frame #8: 0x000ff4a4 hello`(anonymous namespace)::local::MyHeap::collect(vm::Heap::CollectionType, unsigned int, int) + 222 at heap.cpp:1739
    frame #9: 0x000ff3c6 hello`(anonymous namespace)::local::MyHeap::collect(vm::Heap::CollectionType, unsigned int, int) [inlined] (anonymous namespace)::local::collect(c=<unavailable>)::local::Context*) + 464 at heap.cpp:1819
    frame #10: 0x000ff1f6 hello`(anonymous namespace)::local::MyHeap::collect(this=<unavailable>, type=<unavailable>, incomingFootprint=<unavailable>, pendingAllocation=<unavailable>) + 74 at heap.cpp:1989
    frame #11: 0x001bffc0 hello`(anonymous namespace)::doCollect(t=0x1515e200, type=MinorCollection, pendingAllocation=1048600) + 104 at machine.cpp:3456
    frame #12: 0x001bff14 hello`vm::collect(t=0x1515e200, type=MajorCollection, pendingAllocation=1048600) + 92 at machine.cpp:4315
    frame #13: 0x001bfde0 hello`vm::allocate3(t=0x1515e200, allocator=0x14d4a860, type=FixedAllocation, sizeInBytes=1048584, objectMask=false) + 532 at machine.cpp:4259
    frame #14: 0x001c0620 hello`vm::makeByteArray(vm::Thread*, unsigned long) [inlined] vm::allocate2(t=0x1515e200, sizeInBytes=<unavailable>, objectMask=false) + 76 at machine.cpp:4175
    frame #15: 0x001c0612 hello`vm::makeByteArray(vm::Thread*, unsigned long) at machine.h:1639
    frame #16: 0x001c0612 hello`vm::makeByteArray(t=0x1515e200, length=1048576) + 62 at type-constructors.cpp:1206
    frame #17: 0x000e2cce hello`(anonymous namespace)::local::makeBlankArray(t=<unavailable>, type=<unavailable>, length=<unavailable>)::local::MyThread*, unsigned int, int) + 74 at compile.cpp:2563
    frame #18: 0x001fd364 hello`Hello.<init>(J)V + 280
    frame #19: 0x000e1cb4 hello`.LvmInvoke_argumentTest + 16
  frame #20: 0x01d8450f hello`
```

dicej:
In a bootimage=true build, we create allocate certain objects as
"immortal fixies", which means they will never been deallocated at
runtime and should only be visited if/when they point to objects which
might move during garbage collection.  However, there was a bug in the
following case:

 1. immortal fixie F is updated to point to a movable object M and
 thus F is added to the list of fixies to visit during the next minor
 collection (but not the next major one, since all reachable objects
 are visited during a major collection, and there's no point in
 visiting an unreachable object, whereas during a minor collection we
 have to visit F because we don't know if it's reachable or not)

 2. a major collection occurs, but F is not reachable and thus is not
 visited, whereas M is moved

 3. a minor collection occurs, and since F is still in the list, it is
 visited, but since it contains a stale pointer to M's old location,
 we crash

The solution is to ensure unreachable immortal fixies are removed from
the above list after each major collection, thus guaranteeing they
won't be visited on any subsequent collection.

Fixes #443

### DIFF
--- a/src/heap/heap.cpp
+++ b/src/heap/heap.cpp
@@ -1001,8 +1001,8 @@ void sweepFixies(Context* c)
   assertT(c, c->markedFixies == 0);
 
   if (c->mode == Heap::MajorCollection) {
-    free(c, &(c->tenuredFixies));
-    free(c, &(c->dirtyTenuredFixies));
+    free(c, &(c->tenuredFixies), true);
+    free(c, &(c->dirtyTenuredFixies), true);
 
     c->tenuredFixieFootprint = 0;
   }


joshuawarner32:
Is there a test we can add to make sure this stays fixed (perhaps a distilling of #443)?

dicej:
I'll see if I can reproduce it in a way Travis can handle (i.e. without iOS).

dicej:
OK, I can reproduce this using an android/bootimage build on Linux with the following patch:

```
diff --git a/test/Hello.java b/test/Hello.java
index d7d4f91..f4d0551 100644
--- a/test/Hello.java
+++ b/test/Hello.java
@@ -1,5 +1,12 @@
+import java.io.InputStream;
+
 public class Hello {
+    InputStream in = Hello.class.getResourceAsStream("/hello.properties");
+    for (int j = 0; j < 1000; j++) {
+      System.err.println(j);
+      byte[] b = new byte[1024 * 1024];
+    }
     System.out.println("hello, world!");
   }
 }
```

The tricky bit is asking Travis to download and install the Android dependencies.  If we do it the naive way with avian-pack, we'll dramatically increase the time it takes to do a CI run.  I can't think of a way to reproduce this without Android; I'm sure it's possible, but it's not something I want to spend my time on.  Any thoughts?

joshuawarner32:
What's the particular part that requires the android classpath?

dicej:
I think it's something about how the bootimage is build for the Android class library, probably how the static tables (i.e. the singleton objects containing static variables for each class) are allocated.  Other objects allocated this way include system class loaders (i.e. the boot and application loaders), plus addendum objects (those extending avian.Addendum).  I don't know which specific object is triggering the condition, just that it only happens with the Android library.

joshuawarner32:
Hmm... that sounds like it could be a real pain in the ass to isolate.  I'm fine leaving it as is.  Thanks!

PR #446, opened 2015-07-16
Added targets for Eclipse environment descriptor.

bgould:
Someone on the Google Groups site asked how to set up an Eclipse
project with Avian classpath. This patch creates the descriptor
that you can you use to do that at `$(build)/eclipse/jdk/avian.ee`.
The descriptor includes the Avian version, platform, architecture,
and build options to allow for multiple versions to exist side by
side.  Users can import the descriptor into Eclipse via:

    Window >> Preferences >> Java >> Installed JREs >> Add >> Execution
    Environment Description

Once the descriptor is imported, Avian can be used just like any other
JVM installation for Eclipse projects. Personally I use this in
conjunction with Eclim to gain code completion for Avian in vim.

The new targets also create symlinks to loosely mimic OpenJDK's
filenames and folder layout:

    build/linux-x86_64-tails-continuations/eclipse/jdk/
    ├── avian.ee
    ├── bin
    │   └── java -> ../../../avian
    ├── jre
    │   └── lib
    │       └── rt.jar -> ../../../../classpath.jar
    └── src -> ../../classpath

Annoyingly, Eclipse for some reason expects this layout to exist
even though the descriptor format has required parameters for
specifying these locations. I suppose that other software may
look for this "standard" layout in a JVM installation so it may be
generally useful.

These artifacts are only built if the platform is one of `windows`,
`linux`, or `macosx`. The symlinks might not actually work at all on
Windows, I'm not sure how things like cygwin/msys handle that and I
do not have the means to test it. If they do not work a fallback
for windows might be to actually copy the files instead of symlinking.

I realize this can be done outside of the makefile but it seemed
useful to put it here to gain access to the information about the
build location, platform, architecture, and other build options.

For the record, this contribution is my original work and is released
under the same license that Avian uses, found in the license.txt
file in this repository.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1580,11 +1580,32 @@ test-flags = -Djava.library.path=$(build) \
 
 test-args = $(test-flags) $(input)
 
+ifneq ($(filter linux windows macosx,$(platform)),)
+eclipse-exec-env = eclipse-ee
+eclipse-jdk-dir = $(build)/eclipse/jdk
+eclipse-ee-file = $(eclipse-jdk-dir)/avian.ee
+eclipse-bin-dir = $(eclipse-jdk-dir)/bin
+eclipse-lib-dir = $(eclipse-jdk-dir)/jre/lib
+eclipse-src-dir = $(eclipse-jdk-dir)/src
+define eclipse-ee-descriptor
+# An Eclipse execution environment for the Avian JVM\
+\n-Dee.executable=bin/java${exe-suffix}\
+\n-Dee.bootclasspath=jre/lib/rt.jar\
+\n-Dee.language.level=1.7\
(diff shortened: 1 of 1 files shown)


joshuawarner32:
Thanks!

Looks good to me, but I notified the person asking on google groups so they can make sure it works for them.

Also, don't worry about the travis failure.  It's not your fault.  Fixing that is complicated by the fact that it only seems to be reproducible on travis itself.

bgould:
The symlink for the 'src' directory in my original patch was incorrect... I had it pointing to the directory with the compiled .class files.

Updated so that it points to the correct directory with the *.java files.  This makes it so that when you click on a compiled class from classpath.jar in Eclipse, it brings up a read-only view of that source code.

PR #449, opened 2015-07-29
Fixed method signatures on format() methods to return PrintStream as in OpenJDK

bgould:
When I first implemented this I failed to notice that printf on java.io.PrintStream returns a reference to the PrintStream object for chaining method calls.  This causes exceptions at runtime in Avian if you try running code compiled for OpenJDK's standard library, and/or vice versa.  This patch brings the message signatures in line so that these exceptions do not occur.


joshuawarner32:
Thanks!

PR #450, opened 2015-07-29
Closeable extends AutoCloseable; RandomAccessFile implements DataInput

bgould:
In OpenJDK, Closeable extends AutoCloseable so that you can use streams with Java 7's "try with resources" syntax.  Patch make Closeable extend from AutoCloseable to enable this functionality for Avian.

Also, RandomAccessFile in OpenJDK implements the DataInput interface.  Patch adds methods to RandomAccessFile to implement the interface.  I could add tests for this part if you'd like, but I copied the code almost exactly from Avian's DataInputStream and it looked straightforward, so I didn't.  Please let me know if you'd like me to do so.

### DIFF
--- a/classpath/java/io/Closeable.java
+++ b/classpath/java/io/Closeable.java
@@ -10,7 +10,7 @@
 
 package java.io;
 
-public interface Closeable {
+public interface Closeable extends AutoCloseable {
   void close()
     throws IOException;
 }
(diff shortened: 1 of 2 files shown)


joshuawarner32:
From a performance perspective, this is less than ideal, because of the byte-array allocation for every `read{Byte,Short,Int}` call.  However, I can't even tell you with confidence that the allocation pressure and GC-latency will be visible in profiles above the overhead of the JNI call.  If you have a performance-sensitive workload for this, I would suggest making another overload of readBytes that shuffles around up to 8 bytes in a `long`, instead of a byte array.  That's extra code and complexity, though.

That's a really long-winded way of saying, "I think this is fine; let's reassess performance as necessary."

Thanks!

PR #451, opened 2015-08-05
implement basic Java 8 lambda support

dicej:
The two big pieces here are basic invokedynamic support and a working version of LambdaMetaFactory.metafactory.  The latter works by dynamically building a synthetic class with three methods: a static factory method, a constructor for the factory method to call, and a method to satisfy the requested interface which defers to the pecified MethodHandle.
    
This work relies heavily on Avian's specific MethodType and MethodHandle implementations, which provide extra, non-standard features to make code generation easier.  That means we'll probably need to use Avian's versions of java.lang.invoke.* even when building with the OpenJDK or Android class libraries (which is not yet supported, but my next goal).


joshuawarner32:
Nice!

Two small things:
* I think the history could use a little squashing.  The distinct commits I made were nice for debugging, but they're messy and add little value.
* Adding the following to 'travis.yml' might fix the travis build.  Unfortunately, it looks like they may not have support for an 'openjdk8' option.

```
jdk:
  - oraclejdk8
```

dicej:
Yeah, even once we switch Travis to JDK 8, I'm guessing it won't work.  I haven't even tried building against OpenJDK yet.  I'll play with that next.

Do we want to try to support JDK 7 going forward, or shall I just update the documentation to say JDK 8 is the minimum?

dicej:
To clarify the above, I meant I haven't tried building with the openjdk=... or openjdk-src=... options yet.

joshuawarner32:
I was only talking about making travis happy about the current state of the build.  We should either switch to java 8 altogether (specifying it in the docs and the `.travis.yml`), or make compiling the InvokeDynamic test optional (i.e. only if we detect a java 8 compiler).  The latter may be the easier way forward.  We can always switch to the former later, once everything is running stably on 8.

dicej:
Seems like Travis is ignoring the oraclejdk8 bit.  The test wouldn't pass anyway, since I haven't implemented invokedynamic for process=interpret yet.  I'll do that next, plus make the makefile JDK7-compatible.

dicej:
Okay, all tests are passing on Linux with openjdk=$JDK8 and openjdk-src=$JDK8_SRC.  A couple of open questions which could be addressed in future pull requests:

 * How do we convince Travis to use Java 8?

 * How do we handle lambdas in AOT-compiled code?  The current implementation relies on runtime-generated code.  However, unless I'm missing something, there's nothing really dynamic about how lambdas use invokedynamic -- the CallSite returned by LambdaMetafactory.metafactory always points to the same MethodHandle, and there's nothing to prevent that MethodHandle from being generated at build time.  So in theory all we need to do is generate that code ahead of time and replace the invokedynamic with e.g. an invokestatic or invokespecial which calls the code.  Does that make sense?

dicej:
A complication for the AOT case has occurred to me: since the logic to generate the synthetic class for each lambda resides in ```LamdaMetafactory.metafactory```, which is Java code, and there's currently no mechanism for executing Java code in the bootimage generator (the compiler only knows how to generate code for the target, not how to generate and run it for the host), we'll either need to reimplement ```LamdaMetafactory.metafactory``` in C++ or add support for executing Java code to the bootimage generator.  I'm inclined to do the latter, but I'm open to other ideas.

joshuawarner32:
The complications I see with adding support for executing Java code in the bootimage-generator:
* That requires allowing simultaneous `Process`es to be linked in - either an interpreter and a compiler, or two potentially different compilers (when cross-compiling).  At the very least, they will differ by whether they're emitting position independent code or not.
* In the latter case of two compilers, that also means we have to support two different simultaneous x86 assemblers (for instance) - one x86 and one x86_64.  Similarly for arm.

Those would all be good things to support, IMO, but they are substantial pieces of work.

It also occurs to me that we could _cheat_ by running LambdaMetafactory in a different process from the bootimage-generator, either forked and communicating, or as a pre-processing step.

Matter-of-fact, there's already a project for delambdafying Java code, to run on older VMs: https://github.com/orfjackal/retrolambda.

invoke.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
joshuawarner32: This file probably doesn't belong in here.  I wrote this as a quick hack to extract the java/lang/invoke API from metadata.  Even then, there were plenty of manual fixup steps after running it, so it's not readily reusable.

joshuawarner32:
Sorry, I didn't notice you had updated the PR.  If any @github folk happen across this: notifications when changes are pushed would be *really* nice.

Thanks!

dicej:
Sorry, I probably should have said something.

I'm in Mexico now and don't plan to do any work until I return, but here's
my plan FYI:

I think being able to run Java code in the bootimage generator would be
handy, both for the case we've been discussing and future work.  And it
ought to be pretty simple: just modify the makefile to build a VM for the
host as a shared library which has a simple interface (don't need the whole
Java VM invocation interface) and link that in to the bootimage generator
for the target.  No need to play games with namespaces or complicated
linking rules or anything.  Then we just pass a pointer to that VM to the
code generator which it can use when it needs to run Java code.  They could
communicate via a pair of stdin/stdout-style byte streams or whatever --
something generic and reusable.  They'll be isolated from each other across
the link boundary otherwise.  Does that sound reasonable?

joshuawarner32:
I'd be worried about signals behaving weirdly with two VMs in one process like that - though, if you feel like making sure that works, that'd be a great direction to go.

Otherwise, I'd probably either leave them as separate processes, communicating on real stdin/stdout, with some simple IPC system (it doesn't have to be very general) - or link in an interpreter instead, which should avoid the signals problem.

PR #454, opened 2015-09-13
support AOT-compilation of Java 8 lambda expressions

dicej:
These expressions are tricky because they rely on invokedynamic, which
normally implies runtime code generation.  However, since lambdas
don't actually use the "dynamicness" of invokedynamic, we can convert
them into static calls to synthetic classes at compile time.

Since I had already written code to synthesize such classes in Java
and I didn't want to rewrite it in C++, I needed to add support for
running Java code to the bootimage generator.  And since the primary
VM used by the generator is purpose-built to generate AOT-compiled
code for a specific target architecture and is not capable of
generating or running JIT-compiled code for the host architecture, I
added support for loading a second, independent, host-specific VM for
running Java code.

The rest of the patch handles the fact that each method compilation
might cause new, synthetic classes to be created, so we need to make
sure those classes and their methods are included in the final heap
and code images.  This required breaking some giant code blocks out of
makeCodeImage into their own methods, which makes the diff look
scarier than it really is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,8 +168,8 @@ Library" below for details.
 These flags determine the name of the directory used for the build.
 The name always starts with _${platform}-${arch}_, and each non-default
 build option is appended to the name.  For example, a debug build with
-bootimage enabled on Linux/i386 would be built in
-_build/linux-i386-debug-bootimage_.  This allows you to build with
+bootimage enabled on Linux/x86_64 would be built in
+_build/linux-x86_64-debug-bootimage_.  This allows you to build with
 several different sets of options independently and even
 simultaneously without doing a clean build each time.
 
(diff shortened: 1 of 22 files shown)


dicej:
@joshuawarner32, do you have time to look at this?

joshuawarner32:
Apologies for the delay.  The code looks fine.  Is there a way to add tests for this?

I upgraded the Dockerfiles to openjdk 8, if that helps: https://github.com/ReadyTalk/avian/pull/456.  The updated image is already pushed to docker hub as joshuawarner32/avian-build.

dicej:
One way to test this would be to add a makefile target which builds an AOT-compiled boot image that contains all the tests in the suite and runs them each.  Currently, when you run "make bootimage=true test" the class library classes are AOT-compiled, but the test classes themselves are JIT-compiled.  I think that should still be the default, but we should have a version that AOT-compiles everything also.

I'll work on that when I get a chance; maybe next weekend or the one after that.

joshuawarner32:
Cool.  I don't have any concerns with just merging this now, though, so...

PR #459, opened 2015-10-19
add bootimage-test option to makefile

dicej:
This option specifies that the test classes should be AOT-compiled
along with the class library, which allows us to test that everything
works in AOT-compiled form as well as JIT-compiled form.  This is
primarily motivated by the need to test d906db6 (support for
AOT-compilation of Java 8 lambda expressions).

Note that I had to tweak Misc because it tested something that
couldn't be done in an AOT build without a lot of extra work, and
SystemClassLoader.getPackage because it was returning null when the
requested package could not be populated with JAR manifest metadata.
Technically, we probably *should* return null for packages that don't
exist at all (in the sense that no classes have been loaded from such
a package), but tracking that kind of thing seems like more trouble
than it's worth unless someone complains about it.

### DIFF
--- a/makefile
+++ b/makefile
@@ -45,6 +45,11 @@ ifneq ($(lzma),)
 endif
 ifeq ($(bootimage),true)
 	options := $(options)-bootimage
+	ifeq ($(bootimage-test),true)
+    # this option indicates that we should AOT-compile the test
+    # classes as well as the class library
+		options := $(options)-test
+	endif
 endif
 ifeq ($(tails),true)
 	options := $(options)-tails
@@ -108,6 +113,12 @@ winrt ?= $(root)/winrt
 wp8 ?= $(root)/wp8
 
 classpath = avian
+
+bootimage-classpath = $(classpath-build)
+
+ifeq ($(bootimage-test),true)
+	bootimage-classpath = $(classpath-build):$(test-build)
+endif
 
 test-executable = $(shell pwd)/$(executable)
 boot-classpath = $(classpath-build)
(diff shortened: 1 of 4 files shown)


makefile
```diff
@@ -109,6 +114,12 @@ wp8 ?= $(root)/wp8
 
 classpath = avian
 
+bootimage-classpath = $(classpath-build)
+
+ifeq ($(bootimage-test),true)
```
joshuawarner32: To be consistent with the adjustment of `options` above, we might want to wrap this in a `ifeq ($(bootimage),true)`.  Otherwise, you could accidentally build with bootimage-test, but not bootimage and get unexpected results.

Better yet, maybe add a check to the parameter validation code.

dicej: Yeah, I wasn't sure exactly what kind of command line interface I wanted
for this, and I just went for what took the least amount of effort.  The
good news is that if you specify bootimage-test=true without
bootimage=true, nothing unusual will happen, since bootimage-classpath will
simply go unused.

I guess we could either set bootimage=true automatically if
bootimage-test=true is specified, or else complain that the latter was
specified without the former.  Which would you prefer?

joshuawarner32: My initial reaction is to complain.

dicej: "My initial reaction is to complain."  Words to live by.

joshuawarner32: Well said, sir!

joshuawarner32:
Alas, the trace test failed before travis could get to `bootimage-test=true`

PR #464, opened 2015-11-01
- fixed building for iOS 9

benstadin:
- handle sim flag to better distinguish between ios simulator / ios
- added ios_deployment_target to set ios min version
- default to 64 bit build for iOS

### DIFF
--- a/makefile
+++ b/makefile
@@ -763,13 +763,14 @@ ifeq ($(kernel),darwin)
 	rpath =
 
 	ifeq ($(platform),ios)
-		ifeq (,$(filter arm arm64,$(arch)))
+		ifeq ($(sim),true)
 			target = iPhoneSimulator
 			sdk = iphonesimulator$(ios-version)
 			ifeq ($(arch),i386)
 				arch-flag = -arch i386
 			else
 				arch-flag = -arch x86_64
+				arch = x86_64
 			endif
 			release = Release-iphonesimulator
 		else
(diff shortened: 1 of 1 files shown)


benstadin:
I figured I didn't remove the default copyright notices that Xcode creates for each source file. Shall I remove them?

dicej:
Hi Ben,

Thanks for doing this and updating the hello-ios project.  I assume you meant to ask the copyright notice question on the hello-ios PR rather than this one.  Anyway, yes, we should probably remove those so they don't cause confusion.

I suppose we should really replace them with an appropriate license notice, but I never took the time to do that with the original code, so just removing them entirely is fine for now.  The intention is that it's public domain example code.

benstadin:
Hi Joel,

I've removed the copyright notes and edited the pull request. I explicitly mention in the commit message that the code is public domain.


Von: Joel Dice <notifications@github.com<mailto:notifications@github.com>>
Antworten an: ReadyTalk/avian <reply@reply.github.com<mailto:reply@reply.github.com>>
Datum: Montag, 2. November 2015 00:41
An: ReadyTalk/avian <avian@noreply.github.com<mailto:avian@noreply.github.com>>
Cc: Benjamin Stadin <benjamin.stadin@heidelberg-mobil.com<mailto:benjamin.stadin@heidelberg-mobil.com>>
Betreff: Re: [avian] - fixed building for iOS 9 (#464)


Hi Ben,

Thanks for doing this and updating the hello-ios project. I assume you meant to ask the copyright notice question on the hello-ios PR rather than this one. Anyway, yes, we should probably remove those so they don't cause confusion.

I suppose we should really replace them with an appropriate license notice, but I never took the time to do that with the original code, so just removing them entirely is fine for now. The intention is that it's public domain example code.

-

dicej:
Thanks.  I'm updating my Xcode now so I can test this myself.

PR #468, opened 2015-12-18
HTTP URL connection support added and fixes to Socket and BufferedInputStream

keinhaar:
new URL("http://www.XYZ.abc/here/we/go.txt").openStream() 
will no longer result in an UnsupportedOperationException.

### DIFF
--- a/classpath/avian/http/Handler.java
+++ b/classpath/avian/http/Handler.java
@@ -10,17 +10,111 @@
 
 package avian.http;
 
-import java.net.URL;
-import java.net.URLStreamHandler;
-import java.net.URLConnection;
+import java.io.BufferedInputStream;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.FileNotFoundException;
-import java.io.File;
-import java.io.FileInputStream;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.net.Socket;
+import java.net.URL;
+import java.net.URLConnection;
+import java.net.URLStreamHandler;
+import java.util.HashMap;
+import java.util.Map;
 
-public class Handler extends URLStreamHandler {
-  protected URLConnection openConnection(URL url) {
-    throw new UnsupportedOperationException();
-  }
+public class Handler extends URLStreamHandler
+{
+    public URLConnection openConnection(URL url) throws IOException
+    {
+        return new HttpURLConnection(url);
+    }
+    
+    class HttpURLConnection extends URLConnection
+    {
+        Socket socket;
+        private BufferedWriter writer;
+        private InputStream bin;
+        private Map<String,String> header = new HashMap<String, String>();
+        private int status;
+        
+        protected HttpURLConnection(URL url)
+        {
+            super(url);
+        }
+
+        @Override
+        public void connect() throws IOException
+        {
+            if(socket == null)
+            {
+                URLConnection con = null;
+                String host = url.getHost();
+                int port =url.getPort();
+                if(port < 0) port = 80;
+                socket = new Socket(host, port);
+                OutputStream out = socket.getOutputStream();
+                writer = new BufferedWriter(new OutputStreamWriter(out));
+                writer.write("GET " + url.getPath() + " HTTP/1.1");
+                writer.write("\r\nHost: " + host);
+                writer.write("\r\n\r\n");
+                writer.flush();
+                bin = new BufferedInputStream(socket.getInputStream());
+                readHeader();
+//                System.out.println("Status: " + status);
+//                System.out.println("Headers: " + header);
+            }
+        }
+
+        private void readHeader() throws IOException
+        {
+            byte[] buf = new byte[8192];
+            int b = 0;
+            int index = 0;
+            while(b >= 0)
+            {
+                if(index >= 4 && buf[index-4] == '\r' && buf[index-3] == '\n' && buf[index-2] == '\r' && buf[index-1] == '\n')
+                {
+                    break;
+                }
+                b = bin.read();
+                buf[index] = (byte) b;
+                index++;
+                if(index >= buf.length)
+                {
+                    throw new IOException("Header exceeded maximum size of 8k.");
+                }
+            }
+            BufferedReader reader = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(buf, 0, index)));
+            String line = reader.readLine();
+            int x = line.indexOf(' ');
+            status = Integer.parseInt(line.substring(x + 1 , line.indexOf(' ', x+1)));
+            while(line != null)
+            {
+                int i = line.indexOf(':');
+                if(i > 0)
+                {
+                    header.put(line.substring(0, i), line.substring(i + 1) .trim());
+                }
+                line = reader.readLine();
+            }
+            reader.close();
+        }
+
+        @Override
+        public InputStream getInputStream() throws IOException
+        {
+            connect();
+            return bin;
+        }
+        
+        @Override
+        public OutputStream getOutputStream() throws IOException
+        {
+            throw new UnsupportedOperationException("Can' write to HTTP Connection");
+        }
+    }
 }
--- a/classpath/java/io/BufferedInputStream.java
+++ b/classpath/java/io/BufferedInputStream.java
@@ -25,17 +25,16 @@ public class BufferedInputStream extends InputStream {
     this(in, 4096);
   }
 
-  private void fill() throws IOException {
+  private int fill() throws IOException {
     position = 0;
     limit = in.read(buffer);
+
+    return limit;
   }
 
   public int read() throws IOException {
-    if (position >= limit) {
-      fill();
-      if (limit == -1) {
-        return -1;
-      }
+    if (position >= limit && fill() == -1) {
+      return -1;
     }
 
     return buffer[position++] & 0xFF;
@@ -43,7 +42,9 @@ public class BufferedInputStream extends InputStream {
 
   public int read(byte[] b, int offset, int length) throws IOException {
     int count = 0;
-
+    if (position >= limit && fill() == -1) {
+      return -1;
+    }
     if (position < limit) {
       int remaining = limit - position;
       if (remaining > length) {
@@ -54,28 +55,7 @@ public class BufferedInputStream extends InputStream {
 
       count += remaining;
       position += remaining;
-      offset += remaining;
-      length -= remaining;
     }
-
-    while (length > 0) {
-      int c = in.read(b, offset, length);
-      if (c == -1) {
-        if (count == 0) {
-          count = -1;
-        }
-        break;
-      } else {
-        offset += c;
-        count += c;
-        length -= c;
-
-        if (in.available() <= 0) {
-          break;
-        }
-      }
-    }
-
     return count;
   }
 
--- a/classpath/java/net/Socket.java
+++ b/classpath/java/net/Socket.java
@@ -86,6 +86,7 @@ public class Socket implements Closeable, AutoCloseable {
 		
 		@Override
 		public int read(byte[] buffer) throws IOException {
+			if(buffer.length == 0) return 0; //spec says return 0 if buffer length is zero.
 			int fullSize = buffer.length;
 			int index = 0;
 			int size;
@@ -93,11 +94,11 @@ public class Socket implements Closeable, AutoCloseable {
 				size = recv(sock, buffer, index, Math.min(fullSize, Socket.BUFFER_SIZE));
 				fullSize -= size;
 				index += size;
-			} while (fullSize != 0 && size != 0);
+			} while (fullSize != 0 && size > 0); //changed size != 0 to size > 0, because recv returns -1 on error, which should cancel the read, and throw an Exception.
+			if(size < 0) throw new IOException("Error while reading stream"); //as the manpage of recv says, a value below zero indicates an error.
+			if(index == 0 && size == 0) return -1; // if no bytes have been read until now, and the stream is closed (size == 0), then return -1 to indicate end of stream. 
 			return index;
 		}
-
-		
 	}
 	
 	private class SocketOutputStream extends OutputStream {
--- a/test/SocketTest.java
+++ b/test/SocketTest.java
@@ -1,0 +1,23 @@
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.Socket;
+
+public class SocketTest
+{
+
+    public static void main(String[] args) throws IOException
+    {
+        Socket socket = new Socket("mail.gmx.de", 25);
+        InputStream in = socket.getInputStream();
+        OutputStream out = socket.getOutputStream();
+        byte[] buf = new byte[65536];
+        int c = in.read(buf);
+        System.out.println(new String(buf,0,c));
+        out.write("HELO a\r\n".getBytes());
+        out.flush();
+        c = in.read(buf); 
+	if(c < 0) throw new RuntimeException("Answer should be '250 OK'");
+    }
+}


classpath/java/io/BufferedInputStream.java
```diff
@@ -22,7 +22,7 @@ public BufferedInputStream(InputStream in, int size) {
   }
   
   public BufferedInputStream(InputStream in) {
-    this(in, 4096);
+    this(in, 65536);
```
joshuawarner32: What's the reason for bumping this?

classpath/java/net/Socket.java
```diff
@@ -93,7 +93,9 @@ public int read(byte[] buffer) throws IOException {
 				size = recv(sock, buffer, index, Math.min(fullSize, Socket.BUFFER_SIZE));
 				fullSize -= size;
 				index += size;
-			} while (fullSize != 0 && size != 0);
+			} while (fullSize != 0 && size > 0);
+			if(size < 0) throw new IOException("Error whilereading stream");
```
joshuawarner32: Nit-pick: missing space in message.

classpath/java/net/Socket.java
```diff
@@ -93,7 +93,9 @@ public int read(byte[] buffer) throws IOException {
...
+			if(index == 0 && size == 0) return -1;
```
joshuawarner32: There's the edge case here where `buffer.length == 0`.  Intuitively (and according to the spec: http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[]) ) - but if I'm reading this right, this code will return -1 in that case.

classpath/java/io/BufferedInputStream.java
```diff
@@ -43,7 +43,12 @@ public int read() throws IOException {
 
   public int read(byte[] b, int offset, int length) throws IOException {
     int count = 0;
-
+    if (position >= limit) {
+	fill();
+      if (limit == -1) {
+        return -1;
+      }
+	}
```
joshuawarner32: Nit-pick: weird indentation.

joshuawarner32: Also, this block could be simplified to:

```
if (fill() == -1) {
  return -1;
}
```

classpath/java/io/BufferedInputStream.java
```diff
@@ -54,28 +59,7 @@ public int read(byte[] b, int offset, int length) throws IOException {
 
       count += remaining;
       position += remaining;
-      offset += remaining;
-      length -= remaining;
     }
-
-    while (length > 0) {
```
joshuawarner32: Without this loop, it seems like this function will never be able to read more that a single buffer's worth of data at a time.  Perhaps that's why you bumped the default size?

That seems less than idea to me.

classpath/avian/http/Handler.java
```diff
@@ -10,17 +10,112 @@
...
+                writer.write("\n\n");
+                writer.flush();
+                bin = new BufferedInputStream(socket.getInputStream());
+                readHeader();
+//                System.out.println("Status: " + status);
+//                System.out.println("Headers: " + header);
+            }
+        }
+
+        private void readHeader() throws IOException
```
joshuawarner32: This implementation looks rather slow and error-prone:

* We probably ought to have a higher limit than 2K on the maximum size of headers (for reference, Apache uses 8K and IIS uses 16K).
* Furthermore, we should probably throw an error if that limit is exceeded, rather than trying to parse the incomplete header block.
* It seems strange to stop the block after any four newline-like characters.  The spec says "\r\n\r\n" should be the terminator, but this would accept "\r\r\r\r", among other oddities.
* Reading headers byte-by-byte off the wire is going to be needlessly slow.  Ideally, this would be implemented by directly parsing from some intermediate buffers, with a state machine to keep track of where we are.  Note that this would require some extra complexity in having to keep track any data we read off the wire but later determined was not part of the header block (i.e. we saw the terminator in the middle of a block; we need to hold on to the rest of it).  We'll need to feed that to the client as the first part of the returned input stream.

classpath/avian/http/Handler.java
```diff
@@ -10,17 +10,112 @@
...
+            {
+                URLConnection con = null;
+                String host = url.getHost();
+                int port =url.getPort();
+                if(port < 0) port = 80;
+                socket = new Socket(host, port);
+                OutputStream out = socket.getOutputStream();
+                writer = new BufferedWriter(new OutputStreamWriter(out));
+                writer.write("GET " + url.getPath() + " HTTP/1.1");
+                writer.write("\nHost: " + host);
```
joshuawarner32: "\r\n", not just "\n".

classpath/avian/http/Handler.java
```diff
@@ -10,17 +10,112 @@
...
+                writer.write("\n\n");
```
joshuawarner32: Again, "\r\n\r\n".

classpath/avian/http/Handler.java
```diff
@@ -10,17 +10,112 @@
...
+        {
+            if(socket == null)
+            {
+                URLConnection con = null;
+                String host = url.getHost();
+                int port =url.getPort();
+                if(port < 0) port = 80;
+                socket = new Socket(host, port);
+                OutputStream out = socket.getOutputStream();
+                writer = new BufferedWriter(new OutputStreamWriter(out));
```
joshuawarner32: IMO, we ought to be doing the utf8 conversion internally.  As it is, this will cause these string to bounce back and forth between utf8 and utf16 a couple times.  "GET " will initially be utf8 (ish) in memory, but concatenating that will cause it to be converted to utf-16, copied between several internal buffers, and finally converted back to utf8 and sent on the wire.

That said, while this won't the the most efficient, it's not the end of the world.

classpath/java/io/BufferedInputStream.java
keinhaar: Because it can boost the performance a lot if the underlying inputstream delivers data fast enough. This is an experience i made some years ago in a project where we transferred gigabytes of data around the network. On slow connections it wastes only a small memory amount, but is still as fast as before.

classpath/java/net/Socket.java
keinhaar: Will change this....

classpath/java/net/Socket.java
keinhaar: You're right, and i will adress this too. But it is working right now with any other buffer size then zero (which, i think, nobody would ever use), which wasn't the case before.

classpath/java/io/BufferedInputStream.java
keinhaar: Copied this from the simple read() method, but i can easily change this too. Btw. The indentation was caused by that crazy windows editor i used. Will now correct it on Linux.

classpath/java/io/BufferedInputStream.java
keinhaar: With that loop the read operation will block until more data is available. Which may take a while. I think it's more efficient to return the available data. Or add a check before trying to read.

classpath/java/io/BufferedInputStream.java
keinhaar: needed to change the fill() return type to do so.

classpath/avian/http/Handler.java
keinhaar: 1. OK, will change the size to 8k.
2. I will add an Exception.
3. makes it more complicated, but if you think such situations may occur in real life...
4. I'm not reading byte by byte! This is why i've used an BufferedInputStream!

classpath/avian/http/Handler.java
keinhaar: I don't know how to avoid this.

classpath/avian/http/Handler.java
```diff
@@ -10,17 +10,112 @@
...
+
+        private void readHeader() throws IOException
+        {
+            byte[] buf = new byte[2048];
+            int b = 0;
+            int newLineCount = 0;
+            int index = 0;
+            while(b >= 0 && newLineCount != 4)
+            {
+                b = bin.read();
```
joshuawarner32: This is where you're reading byte by byte. I'm pretty sure that will do a jni call for every individual byte, which is expensive.

classpath/avian/http/Handler.java
joshuawarner32: The idea is rather that write constant strings to a Writer, convert them ahead of time to UTF8 byte arrays with .getBytes() - and write them directly to the output stream.  That way, you'll only be doing one UTF8 conversion on each request - the irk path itself.

classpath/java/io/BufferedInputStream.java
joshuawarner32: Oh, wait, sorry.  For some reason I was reading the implementation of the single-byte read() above as fill.  (facepalm).

With the _real_ fill implementation in mind, this is probably the best you can do.

classpath/java/io/BufferedInputStream.java
joshuawarner32: Careful, because that's a deviation from the spec: "This method blocks until input data is available" (http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[])).

Agreed - that may not be the most efficient (for certain scenarios, anyway) - but it is what we need to be doing here.

If you're set on using non-blocking communication (which, by the way, is not a panacea, for either performance or code complexity), you should switch to using the java.nio.channels.SocketChannel, which is non-blocking.

classpath/java/net/Socket.java
joshuawarner32: Could you provide a test-case that demonstrates the bug you found in this code?  That'd make understanding the changes much easier.

classpath/java/io/BufferedInputStream.java
joshuawarner32: Maybe, but it would take a lot to convince me that this is a better default for every possible use case (or even better on average).  For instance, if the file you're reading is small (say, 10KB), it's a complete waste of space to allocate a 64KB buffer.  Also keep in mind that, for blocking connections (which this is intended for, as discussed below), this will have to block in order to fill up the buffer.  A bigger buffer means longer blocking and therefore more latency.

If you ever run into your gigabytes-of-data use case again, you can always specify the larger buffer size by calling the other constructor.

classpath/avian/http/Handler.java
joshuawarner32: With regard to point 3, once upon a time, when implementing an internet standard, the prevailing wisdom was to be liberal in what you accept and strict/conservative in what you produce.  That backfired because people NOT following the spec (i.e. just trying to get things working) introduced subtle non-compliance into their code, simply because it worked with the (liberal) endpoint they happened to be using at the time.  Thus, the internet is littered with all sorts of bizarre almost-correct implementations of various protocols.  IMO, that principle turned out to be bad for the long-term sanity of the internet.

Given that, there are two opposing arguments to be made here:

* That we ought to implement something that doesn't take any bullshit, in an effort to put a little pressure on other people implementing the protocol to make sure it's correct.
* That we ought to cave, and implement something like what you have here (i.e. be liberal in what we accept).  If we went down this path, I would say we should be allowing either '\r\n\r\n' (correct) and '\n\n' (a common mistake) as the terminator.

I would opt for the former, as long as we don't immediately see any compatibility problems.

However, that argument is largely academic at this point.  I just wanted to give a little background.

The much more important problem to fix is the byte-by-byte reading, as I pointed out below.

classpath/avian/http/Handler.java
keinhaar: No, because it is a BufferedInputStream i'm reading from, this is just an access to the byte buffer already filled with the first call to read() in BufferedInputStream.

classpath/java/net/Socket.java
joshuawarner32: I would also like to point out that there is a very common pattern where you can run into zero-length reads, if you're using a length-delimited protocol (say, protobuf, for instance):

* Read the length off the wire (say, of a string you're trying to read)
* Allocate a byte array of the appropriate size
* Pass that to `read(...)` to read the appropriate number of bytes.

If you don't gracefully support zero-length reads, that's a bug waiting to happen.  I ran into that exact issue not long ago with a Rust decompression library.

From a security perspective, that's a very dangerous bug to have, since that's an edge case that not many people would have considered (when writing the code).  All sorts of weird stuff could fall out of that.

classpath/avian/http/Handler.java
joshuawarner32: You're right, my bad.

However, you are buffering it three times (once for the first BufferedInputStream, again for the byte array you're reading it into, and once again for the BufferedReader you're using).  That's not the most efficient, but not the end of the world.

joshuawarner32:
@keinhaar - thanks for the PR, and sorry for being so picky.

Here's a list of the important items (IMO) that need to be done before merging:

* Add tests (i.e. .java files in the test directory) that demonstrate the previous bugs, and therefore prove beyond a shadow of a doubt they're fixed.
* Make sure we haven't regressed in compatibility with openjdk, with respect to the BufferedInputStream behavior (as I suspect this code has).  In my mind, the bar is pretty high - we need to make sure that any code written expecting the openjdk behavior (whatever that is), will also work fine on Avian.  I'd like to either revert to the previous code and attempt a smaller fix for whatever bug you found (that test case is important!) - or demonstrate solidly that we're not regressing anything.  The only way I can think of to approach the latter choice is by rigorous automated testing, which will be quite tricky.
* Change the default buffer size back, or provide a very solid argument that the new buffer size isn't going to make things worse for anybody.

All the other comments I made are up for discussion, I think.

Once everything else is finalized, I'd prefer you squashed all the commits into one.  You can see a good tutorial on how to do that here: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/.  That keeps the history clean, and can make things like bisecting easier later.

classpath/java/io/BufferedInputStream.java
keinhaar: As you say... The spec says: "This method blocks until input data is available"! And there IS data available in the buffer. If the buffer is empty the method may still block while filling the buffer. 
Also the current method does not even try to fill the buffer if "read(byte[] b, int offset, int length)" is called. This is very bad if the given buffer "b" is small. This could end up in reading single bytes from the network, if b is of size 1, which would clearly make the BufferedInputStream useless.

joshuawarner32: Let me see if I can summarize:

You're pointing out (rightly) that the code is not well optimized for when `length < this.limit` (perhaps much less, as in your example of a single-byte read).  You're code also seems to handle that better.  On the other hand, the current code _does_ deal well with the case where `length > this.limit`, but unless I'm missing something, your code doesn't handle that well, since it forces a fixed upper limit on the amount of data that can be read per call to `read`.  I think both cases are important.

Does that make sense?  Is that what you're getting at?  If so, perhaps we can come up with a solution that solves both cases well.

keinhaar:
OK, trying to fulfill the important stuff.
1. I changed the buffer size back to 4096, because you're right - i can use the constructor argument to set it in my code.
2. BufferedInputStreams error is just, that it does not match the spec, as i mentioned in the comment above. Don't know how i should write a test, that demonstrates the blocking that occurs even if data is available in the buffer.
3. I will try to demonstrate the Socket bug.
4. I will try to squash the commits. Never done this so far.

Currently i found, that all tests on windows pass, even if i break them intentionally by throwing an exception. On Linux the same tests will fail. Don't know how to fix that. Could you please check that?

joshuawarner32:
Writing a test case will likely involve making a mock implementation of an InputStream, which you can control absolutely from within the test (i.e. tell it how much data to send to the caller at each step, etc).

For example, you might have test code that looks vaguely like the following:

```
MockInputStream is = new MockInputStream(); // an inner class impl of InputStream that returns bogus data (say, zeroes) whenever asked, but records all the `read` calls made
BufferedInputStream b = new BufferedInputStream(is, 10); // tiny buffer, to make testing easier
for(int i = 0; i < 10; i++) {
  expect(b.read(new byte[1]) == 1);
}
expect(is.callsToRead == 1);
```

But keep in mind that test also has to pass when run directly on the system jre, so can't be too nit-picky about implementation-specific behavior.  It's a tricky balance.

---

Can you clarify which test(s) fail for you on linux?  (Note, they can't be failing on _all_ linux systems, since we have Travis continuously checking that - the "All checks have passed" above).  Also, the output those test(s) are giving would be helpful.  You can get that by either looking at build/<config>/log.txt, which contains the concatenated output of all the tests, or you can run `make run input=Sockets` (for example) to run just one test and see the output.

keinhaar:
I made the Enums.java test fail by changing the first line in the main method from "== 0" to "== 1". Just to see how tests are written for avian. Then i ran the tests by invoking "make test", which caused the Enums to fail as i expected (on Linux). Then i did the same on windows, where it passed. 
make run input=Enums provides an error on Linux AND Windows. 
Maybe using "make test" is wrong??

My Testcase for the Socket (SocketInputStream) did not work, because ServerSocket is not implemented in avian and the only other way i could think of to show the error would be to download a file from somewhere. So instead i've now commented all the changes in that class to clearify the reasons. 
By the way, it is blocking until the buffer is filled completely, which will be a problem if you try to setup something like smtp with its "HELO" -> "OK", "MAIL" -> "OK" traffic. Like BufferedInputStream it should return any bytes received even if it is only 1 byte.

joshuawarner32:
Can you post the build/<config>/log.txt?  That may give us some details about what's going wrong.  Also, what environment are you using on windows?  Cygwin?  MSYS?

On the test case, we really ought to find a way to run a server locally.  For example, avian _does_ have java.nio.ServerSocketChannel, which is the nonblocking equivalent.

When you're talking about the SMTP example, my first reaction is that you probably don't want to be using a BufferedAnything anyway.  If you want fine control over how and when input is buffered and delivered, you really need to just take control of that, and get rid of the generic intermediate buffer.  Buffering is generally not the answer in situations where you need low-latency.  I'd consider the effective deadlock you're talking about waiting for a full buffer's worth of response to be an extreme example of high-latency.

However - it looks like Travis failed the build after you added that test in the `jdk-test` step - which indicates to me that your test doesn't even work on openjdk, which means there's a problem with the test, not necessarily with avian.  I'm not super familiar with SMTP, but it seems a little fishy that you'd be waiting for a response from the server (line 16) before sending anything yourself (line 18/19).  Things don't seem to improve even if I swap those around, so I'm not sure what's going on.

keinhaar:
Here's the log.txt from my Windows 7. I'm using MSYS.
[log.txt](https://github.com/ReadyTalk/avian/files/67750/log.txt)

keinhaar:
OK, after another 6 hours of trying to get the Testcase for Sockets working in avian, i'm giving up being frustated.
The test runs fine in Eclipse using JDK, but with avian it blocks for some reason, that i don't understand, or, if run with "make jdk-test" tells me that package "java.net" is not permitted!?
I will close this pull request now, and create a new one with my changes as discussed above. It's working for me, and i can even do SMTP like protocolls (see SocketTest) with or without BufferedInputStream.

keinhaar:
[SocketTest.zip](https://github.com/ReadyTalk/avian/files/67784/SocketTest.zip)

PR #469, opened 2015-12-20
Added support for HTTP URL connections, and fixed SocketInputStream and BufferedInputStream.

keinhaar:
Did some cleanup after initial commit:
- Bigger HTTP Header Buffer
- Exception if Header is anyway exceeded.
- Linebreaks on HTTP Request fixed to standard.
- Only stop header reading on \r\n\r\n and no longer on \n\n\n\n
- Simplyfied the code to stop if buffer could not be filled.
- Handle special case if buffer has length 0, like specified in the Java API
- Socket will no longer fill the buffer completely, because that breaks interactive Protocols like SMTP

### DIFF
--- a/classpath/java/io/BufferedInputStream.java
+++ b/classpath/java/io/BufferedInputStream.java
@@ -7,8 +7,10 @@
 
    There is NO WARRANTY for this software.  See license.txt for
    details. */
-
 package java.io;
+
+import java.io.IOException;
+import java.io.InputStream;
 
 public class BufferedInputStream extends InputStream {
   private final InputStream in;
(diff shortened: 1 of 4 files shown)


classpath/java/io/BufferedInputStream.java
```diff
@@ -54,28 +55,7 @@ public int read(byte[] b, int offset, int length) throws IOException {
 
       count += remaining;
       position += remaining;
-      offset += remaining;
-      length -= remaining;
     }
-
-    while (length > 0) {
```
joshuawarner32: I still think we should leave this while loop in there. Glancing at the openjdk source, they're doing something largely equivalent.  Keep in mind that the `if (in.available() <= 0)` ought to take care of your concern that we'll block on waiting for more data.

keinhaar: OK, what about a compromise? I've written a Test that blocks with avians current BufferedInputStream but runs perfectly well with Oracles JDK. So if you want to keep the loop, lets change it by moving the available() call into the while...
```
    while (length > 0 && in.available() > 0) {
      int c = in.read(b, offset, length);
      if (c == -1) {
        if (count == 0) {
          count = -1;
        }
        break;
      } else {
        offset += c;
        count += c;
        length -= c;
      }
    }
```
This way we still try to fill the given byte array as much as possible, but it will also match the spec and will be compatible to JDK.

joshuawarner32: Yeah, that seems reasonable.

keinhaar:
OK, Testcase commited, and loop readded to BufferedInputStream as suggested in my last post.
This time i was also able to squash the changes into one commit.

joshuawarner32:
Thanks @keinhaar!

PR #470, opened 2015-12-22
Change Avian src volume path to match Dockerfile

Xtansia:
Dockerfile declares the Avian src volume path as /var/src/avian
Updated build.sh to use the same volume path


joshuawarner32:
Thanks!

PR #471, opened 2016-01-10
Added missing apis to ByteBuffer

lwahlmeier:
added float and double getters and puts.
Added isReadOnly() to Buffers, ByteBuffers now set that state in Buffers.
Added duplicate()


lwahlmeier:
Fairly certain that build failure has nothing to do with these changes since it was in the trace tests.  Is there a way to retrigger the build?

joshuawarner32:
Thanks!

Indeed, that's a failure we've seen off-and-on, and so far not been able to reproduce outside of Travis.  In the past I've suggested retriggering the build by doing a `git commit --amend --no-edit; git push -f` - but there might be a better way.

PR #472, opened 2016-01-13
fix openjdk/openjdk-src builds for latest JDKs 7 and 8

dicej:
The tower of patches and hacks grows higher.  Ideally, we'll just drop
support for JDK 7 soon and clean this mess up a bit, but TravisCI
still hasn't gotten the memo that it's dead, so we muddle onward.

I've tested this on Windows, but not yet Linux or OS X.  Wanted to get
a PR before I move on to that.

### DIFF
--- a/makefile
+++ b/makefile
@@ -166,6 +166,10 @@ library-path = $(library-path-variable)=$(build)
 
 
 ifneq ($(openjdk),)
+	openjdk-version := $(shell $(openjdk)/bin/java -version 2>&1 \
+		| head -n 1 \
+		| sed 's/.*version "1.\([^.]*\).*/\1/')
+
 	openjdk-arch = $(arch)
 	ifeq ($(arch),x86_64)
 		openjdk-arch = amd64
@@ -832,12 +836,12 @@ ifeq ($(kernel),darwin)
 		cflags += $(flags)
 		asmflags += $(flags)
 		lflags += $(flags)
-		
+
 		ios-version-min=$(ios-version)
 		ifdef ios_deployment_target
 			ios-version-min=ios_deployment_target
 		endif
-		
+
 		ifeq ($(sim),true)
 			ifeq ($(arch),x86_64)
 				classpath-extra-cflags += \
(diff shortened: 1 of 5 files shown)


dicej:
I've tested this on Mac, Linux, and Windows.  Everything looks fine except on Mac, where the openjdk-src build works fine, but the openjdk build gives:

dyld: Symbol not found: __cg_jpeg_resync_to_restart
  Referenced from: /System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libTIFF.dylib
  Expected in: /Users/dicej/trash/openjdk-1.7.0-u80-unofficial-macosx-x86_64-image/jre/lib/libJPEG.dylib
 in /System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libTIFF.dylib

I've run into this before, so it isn't a new regression.  Apparently the system libTIFF is trying to use the JRE's libJPEG and I'm not sure how to prevent that.  Probably some dlopen magic, but I'm not ready to pursue it now, and maybe it's only a problem on my machine anyway.

Also, I can't find any pre-built OpenJDK 8 images, and building it from source requires downgrading to Xcode 4, which sounds like a nightmare.  Someone who really cares about Avian/OpenJDK 8 on Mac can do that.

joshuawarner32:
Do you know where the __try and __except statements are used in the openjdk code?  I'm worried that there's an interesting case that we're missing in our test suite where a windows structured exception will be thrown in such a block, not get caught, and lead to bizarre bugs down the road.

dicej:
java_props_md.c

I do agree that my preprocessor hack is dangerous, so we could change it to a patch instead.  Short using a compiler that actually supports SEH, I don't know what a proper patch would look like.

joshuawarner32:
Technically, I think we could simulate __try/__except with the preprocessor and windows API calls (the same ones we use in windows/signal.cpp, IIRC), but that may be premature.  I'd definitely be more comfortable with calling out each of the locations explicitly in a patch, since that'll give a more visible breadcrumb to investigate, should we ever run into troubles with this.

dicej:
I've pushed a version that uses a patch instead of the preprocessor.  It's even uglier, but safer.

joshuawarner32:
Looks like the Trace test failed again.  It's gotten obnoxious enough that I'm tempted to do something like just disabling it on PR builds.  What do you think?

dicej:
Works for me.

PR #475, opened 2016-01-16
synchronize on Java class rather than VM class in static synchronized…

### LINKED ISSUE #474: OkHttp throws IllegalMonitorStateException
Hi,

I am trying to use [OkHttp](http://square.github.io/okhttp) with avian but it throws IllegalMonitorStateException. It works perfectly with OkHttp 2.4.0 or lower version but it fails for 2.5.0+ versions.

Avian: compiled from the master branch
openjdk: openjdk 8u60

```java
package test;

import java.io.IOException;

import com.squareup.okhttp.OkHttpClient;
import com.squareup.okhttp.Request;

public class OkHttpTest {

	public static void main(String[] args) throws IOException {
		OkHttpClient client = new OkHttpClient();
		Request request = new Request.Builder().url("http://stackoverflow.com").get().build();
		System.out.println(client.newCall(request).execute().isSuccessful());
	}
}
```
```
java/lang/IllegalMonitorStateException
  at java/lang/Object.notify (native)
  at okio/AsyncTimeout.scheduleTimeout (line 100)
  at okio/AsyncTimeout.enter (line 69)
  at okio/AsyncTimeout$1.write (line 153)
  at okio/RealBufferedSink.flush (line 221)
  at com/squareup/okhttp/internal/http/HttpConnection.flush (line 141)
  at com/squareup/okhttp/internal/http/HttpTransport.finishRequest (line 52)
  at com/squareup/okhttp/internal/http/HttpEngine.readNetworkResponse (line 904)
  at com/squareup/okhttp/internal/http/HttpEngine.access$300 (line 92)
  at com/squareup/okhttp/internal/http/HttpEngine$NetworkInterceptorChain.proceed (line 891)
java/lang/IllegalMonitorStateException  at com/squareup/okhttp/internal/http/HttpEngine.readResponse (line 749)
  at com/squareup/okhttp/Call.getResponse   at java/lang/Object.wait (line 268)
(native)
  at com/squareup/okhttp/Call$ApplicationInterceptorChain.proceed   at java/lang/Object.wait (line 224)
(line 460)
  at com/squareup/okhttp/Call.getResponseWithInterceptorChain   at okio/AsyncTimeout.awaitTimeout (line 195)
(line 323)
  at com/squareup/okhttp/Call.execute   at okio/AsyncTimeout.access$000 (line 79)
(line 40)
  at test/OkHttpTest.main   at okio/AsyncTimeout$Watchdog.run (line 13)
(line 286)
```

dicej:
… methods

Previously, the following code would throw an IllegalMonitorStateException:

public class Test {
  public static synchronized void main(String[] args) {
    Test.class.notify();
  }
}

The problem stems from the fact that for a long time Avian has had two
representations of a given class: avian.VMClass and java.lang.Class.
It used to be that there was only one, java.lang.Class, but that
didn't play nicely with OpenJDK's class library, so we split it into
two.  Unfortunately, we forgot to update the JIT and interpreter
accordingly, so a static synchronized method would acquire the
avian.VMClass instance, whereas Foo.class.notify() would be invoked on
the java.lang.Class instance.

This commit fixes it.

Fixes #474


joshuawarner32:
The travis build failed on `make bootimage=true bootimage-test=true test`, which I'm able to reproduce locally.

dicej:
Looks good now.

PR #480, opened 2016-04-04
update to iOS 9.2 SDK and define sysroot for iOS

dicej:
Previously, we only seemed to define sysroot for MacOS, not iOS, yet
we reference $(sysroot) in both cases.  This ensures it is defined in
both cases as well.

### DIFF
--- a/makefile
+++ b/makefile
@@ -799,7 +799,8 @@ ifeq ($(kernel),darwin)
 		sdk-dir = $(platform-dir)/Developer/SDKs
 
 		ios-version := $(shell \
-				if test -L $(sdk-dir)/$(target)9.1.sdk; then echo 9.1; \
+				if test -L $(sdk-dir)/$(target)9.2.sdk; then echo 9.2; \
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -799,7 +799,8 @@ ifeq ($(kernel),darwin)
 		sdk-dir = $(platform-dir)/Developer/SDKs
 
 		ios-version := $(shell \
-				if test -L $(sdk-dir)/$(target)9.1.sdk; then echo 9.1; \
+				if test -L $(sdk-dir)/$(target)9.2.sdk; then echo 9.2; \
```
joshuawarner32: This is getting slightly out of hand.  How about something like this:?

`for x in 9.3 9.2 9.1; do if test -d $(sdk-dir)/$(target)$x.sdk -o -L $(sdk-dir)/$(target)$x.sdk; then echo $x; break; fi; done`

dicej: Good idea.  I just pushed an update.

joshuawarner32:
Looks like travis had a little hickup.  Could you force-push a no-op change to re-trigger the build?

PR #482, opened 2016-04-25
drop MSYS build support

dicej:
Per a recent request for help on the Google group, I spent some time
this past weekend trying to debug the MSYS build.  The conclusion I
came to is that MSYS is not worth supporting anymore.  The most recent
release is almost three years old, and I've been unable to find any
variant with a 64-bit-target compiler that will actually install.  If
someone else cares enough about building Avian on MSYS to maintain it
themselves, patches are welcome.

Meanwhile, let's all just use Cygwin.  Perhaps someone will build a
cool package manager on top of WSL and suddenly make both Cygwin and
MSYS obsolete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,12 +21,6 @@ to use forward slashes in the path.
     $ make
     $ build/macosx-x86_64/avian -cp build/macosx-x86_64/test Hello
 
-#### on Windows (MSYS):
-    $ git clone git@github.com:ReadyTalk/win64.git ../win64
-    $ export JAVA_HOME="C:/Program Files/Java/jdk1.7.0_45"
-    $ make
-    $ build/windows-x86_64/avian -cp build/windows-x86_64/test Hello
-
 #### on Windows (Cygwin):
     $ git clone git@github.com:ReadyTalk/win64.git ../win64
     $ export JAVA_HOME="/cygdrive/c/Program Files/Java/jdk1.7.0_45"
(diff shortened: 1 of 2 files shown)


bigfatbrowncat:
That looks like a catastrophy for us (avian-pack) cause we use [this toolchain](http://bigfatbrowncat.github.io/cross-building/) for building.

By the way, maybe you should look at  MSYS2: https://msys2.github.io/

It's finely maintained and cozy. Maybe we should move to it instead of dropping MSYS totally?

dicej:
@bigfatbrowncat, sorry, I forgot you guys use MSYS.  I'll play around with MSYS2 and see how it goes.

dicej:
By the way, @bigfatbrowncat, what is the reason you've chosen MSYS instead of Cygwin?  In my experience Cygwin is much easier to use and better maintained.

dscho:
@dicej think of MSYS2 (and MSYS) as a portable Cygwin: you can have more than one MSYS "installation" side by side.

Also, MSYS2 (and MSYS) are stripped down Cygwins with the intention to build non-Cygwin software. Keep in mind that Cygwin comes with a POSIX emulation layer, the Cygwin runtime. That layer is very convenient when "porting" POSIX software to Windows, especially when using `fork()`. It comes at a hefty performance penalty, though. The speed difference between Cygwin and pure Win32 software can easily be in the 20x-60x range. [Git for Windows](https://git-for-windows.github.io/) (maintained by yours truly) uses MSYS2 for that reason.

Note that it should be easy enough for @bigfatbrowncat to have a look at replacing MSYS support in Avian by MSYS2. (And I'll gladly answer questions, I had to do similar work when weaning Git for Windows off of MSYS and latch it onto MSYS2.)

bigfatbrowncat:
1. MSYS is much smaller
2. MSYS is well integrated with mingw64
3. In Cygwin there is a risk to use the default toolchain for something. It links against GPL components so the result wouldn't possibly be GPL-free.

Maybe some of these arguments are naive or outdated...
It just worked from the very beginning, so we used it for long.

bigfatbrowncat:
By the way - maybe Avian contributors could compose their own toolchain for Windows and publish it here (like I did for my own projects). If it's based on MSYS (or MSYS2) it would be portable, environment-agnostic and we could test all thee recent builds against it. What do you think? (I hope that's better than supporting multiple GNU-based build environments on Windows)

dicej:
Okay, I've opened a new PR, #483, which reverts the makefile part so you can still build with MSYS.  However, I'm still removing the README.md instructions for building with MSYS because I think Cygwin is easier for new users, and I have more experience supporting it.

It's fine if you prefer MSYS, it's just not something I want to officially support.  I'm happy to accept patches to fix MSYS-related bugs, of course.

PR #484, opened 2016-04-28
Added support for MSYS2 with mingw-w64 toolchain

bigfatbrowncat:
I've made some minor changes to support MSYS2 (old MSYS should work as well).

In order to config the environment you should only install MSYS2, run and then use pacman (yeah, it does contain package manager inside, cool!) to install mingw64 toolchain like this:

    pacman -S mingw-w64-x86_64-gcc

Then you close the main shell, open "mingw64_shell.bat" (this is a special msys2 shell for mingw64 environment) and

    cd avian && make

Don't forget to set JAVA_HOME ;)

Please, test it on your site and tell me if you have any troubles. Should I update the building instruction as well?

### DIFF
--- a/makefile
+++ b/makefile
@@ -3,7 +3,7 @@ MAKEFLAGS = -s
 name = avian
 version := $(shell grep version gradle.properties | cut -d'=' -f2)
 
-java-version := $(shell $(JAVA_HOME)/bin/java -version 2>&1 \
+java-version := $(shell "$(JAVA_HOME)/bin/java" -version 2>&1 \
 	| head -n 1 \
 	| sed 's/.*version "1.\([^.]*\).*/\1/')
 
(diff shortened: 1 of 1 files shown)


bigfatbrowncat:
Are you sure that using Cygwin is preferable? It's heavy, contains a risky GPL-libraries that could be linked to the user's project by mistake... I'm not sure that mingw64 is easier to be used under Cygwin. Could you tell your thoughts about all that?

dicej:
For me, it's easier to use than MSYS because it's more Linux-like.  I can
run an SSH server on it (maybe that's possible with MSYS, but I never
figured out how), and it has a lot of the Linux tools I'm used to using.

I'm not worried about accidentally linking to GPL'd libraries.  My Linux
system is full of GPL'd libraries and I've never accidentally linked to any
of them.

Anyway, I don't want to debate about which is better.  I think both Cygwin
and MSYS are great; I just find Cygwin easier to use, and I'd rather not
try to support both of them.  If you prefer MSYS, please feel free to
submit pull requests like this one and help out on the Google group (
https://groups.google.com/forum/#!forum/avian) when people have MSYS
questions.  I'll support Cygwin and you can support MSYS.

bigfatbrowncat:
Should the possibility to use MSYS be noted in README? I could add a couple of lines with a disclaimer that it's not officially supported.

dicej:
Yes, that would be fine.  If people have questions, though, I might send
them to you :)

PR #491, opened 2016-07-01
add license title and capitalize filename

waldyrious:
The title is not strictly required, but it's useful metadata, and part of the recommended license template text (see http://choosealicense.com/licenses/isc/ and https://opensource.org/licenses/isc-license)

As for the filename, lowercase is fine, but since this project already uses an uppercase README as customary, using the same standard for the LICENSE file seemed to make sense.


joshuawarner32:
Thanks!

PR #492, opened 2016-07-01
patch Inet4AddressImpl.c to declare IcmpSendEcho2Ex

dicej:
Apparently the MinGW header files don't declare it, despite it being
part of liblphlapi.a.  More confusingly, it didn't break anything for
64-bit builds because the compiler used an implicit declaration that
matched the link-time symbol name.  Alas, no such luck for 32-bit
builds, since the implicit declaration was not annotated with the @48
stdcall argument stack size, so the build died at link time.


mikehearn:
Surely this isn't right? It patches this definition in on UNIX too, which of course doesn't know about DWORD and other Microsoft things.

dicej:
@mikehearn: you're right.  My fault for not noticing that.

Here's a fix: https://github.com/ReadyTalk/avian/pull/504

PR #495, opened 2016-08-12
Fix ios bootimage build

joshuawarner32:


dicej:
Perhaps we should use $(so-prefix) too while we're at it.

joshuawarner32:
@dicej Good point; fixed.

joshuawarner32:
FYI, with this PR, I'm able to successfully `make platform=ios arch=arm bootimage=true mode=debug`

PR #496, opened 2016-08-24
override write(byte[]) in DeflaterOutputStream

dicej:
Otherwise, it will inherit the version from FilterOutputStream, which
is not consistent with the write(byte[],int,int) or write(int)
implementations.


joshuawarner32:
I would have thought we'd already have a test that catches this...

dicej:
We do have a test for zipoutputstream that on shallow inspection looks
pretty thorough, but it didn't catch this or the zip corruption reported on
the mailing list, which I won't have a chance to investigate until I get
back from vacation next week.

PR #498, opened 2016-09-10
disable Trace.java test when running on Travis-CI

dicej:
Trace.java fails randomly on Travis, but we've never been able to
reproduce or debug the failure elsewhere, so now we skip it if we
detect the suite is running on Travis.

### DIFF
--- a/test/Trace.java
+++ b/test/Trace.java
@@ -63,7 +63,7 @@ public class Trace implements Runnable {
     try {
       for (int i = 0; i < 10000; ++i) {
         test(this);
-        
+
         if (i % 100 == 0) {
           System.out.print("r");
           System.out.flush();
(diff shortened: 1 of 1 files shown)


joshuawarner32:
Ironically, the travis build failed for (afaict) unrelated reasons.

dicej:
I'm going to assume the error is something transient with Travis and try again tomorrow with a dummy push.  If it's still broken then, I'll look at it more closely.

joshuawarner32:
FWIW, I think there's something going wonky with the ubuntu mirror they're using: doing `curl -I http://us-central1.gce.archive.ubuntu.com/ubuntu/pool/main/g/gcc-4.6/gcc-4.6-multilib_4.6.3-1ubuntu5_amd64.deb` make it look like it's going through 13 (!) successive proxies.  That's what the 'Got a single header line over 360 chars' message is about.  Whether that's a red herring is a different matter.

joshuawarner32:
I emailed travis support about the failure.  We'll see what they say.

dicej:
Thanks.  I started investigating it a bit myself, but it looked like a rabbit hole, so I moved on to more pleasant tasks.

joshuawarner32:
FWIW, the headers show in the `curl -I` command above have been cleaned up as well.  I haven't heard a response on that email yet, but I'm betting it's not a coincidence.

PR #499, opened 2016-09-16
print informative message if lambda encountered with no host VM

dicej:
Per a recent bug report in the hello-ios project, we found that
bootimage-generator would abort with no explanation if it encountered
a lambda invocation and the `-hostvm` option was unspecified.  This
commit ensures that a helpful message is printed before exiting.


dicej:
Thanks.  BTW, it looks like Travis got past the apt-get stage this time, which is promising.

PR #502, opened 2016-09-22
Update makefile to support tests of InvokeDynamic also to Java 8

teras:


teras:
Indeed this seems like a good and future safe approach

dicej:
Actually, now that I look at this more closely, I don't think any change is needed at all.  The current code *removes* InvokeDynamic.java from the list test files if the Java version is 7, meaning it will only be run if the version is greater than that (or if it's less than that, but we don't support any versions earlier than 7 anyway).

Unless I'm misunderstanding, your patch does the opposite of what it claims -- it removes the InvokeDynamic test on Java 8 instead of adding it.

teras:
I am sorry then. I thought that it performs these tests only on Java 7 code, since the target version was lower than 8.

dicej:
No problem; the test is confusing.  Thanks for all your efforts to improve the project.

PR #505, opened 2016-09-28
Lambda and primitive fixes

mikehearn:
Couple of fixes:

1) Upgrade invokedynamic support so the altMetafactory can be used. javac sometimes emits calls to altMetafactory when a lambda is serializable or otherwise marked with additional interfaces. The code for this is kind of hacky due to the different sources of type data for when method arguments are present vs when they are not due to use of varargs. Ideally the constant pool would have all the needed type information in it, but as far as I can tell it doesn't.

2) Primitive and array classes need to be "public abstract final" for compatibility with HotSpot. Some software does rely on this, e.g. Kryo.

### DIFF
--- a/classpath/java/lang/invoke/LambdaMetafactory.java
+++ b/classpath/java/lang/invoke/LambdaMetafactory.java
@@ -23,6 +23,10 @@ import avian.Assembler;
 import avian.ConstantPool.PoolEntry;
 import avian.SystemClassLoader;
 
+// To understand what this is all about, please read:
+//
+//   http://cr.openjdk.java.net/~briangoetz/lambda/lambda-translation.html
+
 public class LambdaMetafactory {
   private static int nextNumber = 0;
 
@@ -348,14 +352,45 @@ public class LambdaMetafactory {
 
     int flags = (Integer) args[3];
     boolean serializable = (flags & FLAG_SERIALIZABLE) != 0;
+
+    // Marker interfaces are added to a lambda when they're written like this:
+    //
+    //    Runnable r = (Runnable & Serializable) () -> foo()
+    //
+    // The intersection type in the cast here indicates to the compiler what interfaces 
+    // the generated lambda class should implement. Because a lambda has (by definition)
+    // one method only, it is meaningless for these interfaces to contain anything, thus
+    // they are only allowed to be empty marker interfaces. In practice the Serializable
+    // interface is handled specially and the use of markers is extremely rare. Adding
+    // support would be easy though.
     if ((flags & FLAG_MARKERS) != 0)
-      throw new UnsupportedOperationException("FLAG_MARKERS");
+      throw new UnsupportedOperationException("Marker interfaces on lambdas are not supported on Avian yet. Sorry.");
+
+    // In some cases there is a mismatch between what the JVM type system supports and
+    // what the Java language supports. In other cases the type of a lambda expression
+    // may not perfectly match the functional interface which represents it. Consider the
+    // following case:
+    //
+    //    interface I { void foo(Integer i, String s1, Strings s2) }
+    //    class Foo { static void m(Number i, Object... rest) {} }
+    //    
+    //    I lambda = Foo::m
+    //
+    // This is allowed by the Java language, even though the interface representing the 
+    // lambda specifies three specific arguments and the method implementing the lambda
+    // uses varargs and a different type signature. Behind the scenes the compiler generates
+    // a "bridge" method that does the adaptation.
+    //
+    // You can learn more here: http://www.oracle.com/technetwork/java/jvmls2013heid-2013922.pdf
+    // and here: http://cr.openjdk.java.net/~briangoetz/lambda/lambda-translation.html
     if ((flags & FLAG_BRIDGES) != 0) {
       int bridgeCount = (Integer) args[4];
       if (bridgeCount > 0)
-        throw new UnsupportedOperationException("FLAG_BRIDGES");
+        throw new UnsupportedOperationException("A lambda that requires bridge methods was used, this is not yet supported by Avian. Sorry.");
     }
 
+    // TODO: This is not necessary if the function type interface is already inheriting
+    // from Serializable.
     Class[] interfaces = new Class[serializable ? 1 : 0];
     if (serializable)
       interfaces[0] = java.io.Serializable.class;
--- a/classpath/java/lang/invoke/LambdaMetafactory.java
+++ b/classpath/java/lang/invoke/LambdaMetafactory.java
@@ -25,7 +25,11 @@ import avian.SystemClassLoader;
 
 public class LambdaMetafactory {
   private static int nextNumber = 0;
-  
+
+  public static final int FLAG_SERIALIZABLE = 1;
+  public static final int FLAG_MARKERS = 2;
+  public static final int FLAG_BRIDGES = 4;
+
   private static Class resolveReturnInterface(MethodType type) {
     int index = 1;
     byte[] s = type.spec;
@@ -202,13 +206,15 @@ public class LambdaMetafactory {
                       new MethodHandle(implementationClass,
                                        implementationName,
                                        implementationSpec,
-                                       implementationKind));
+                                       implementationKind),
+                      emptyInterfaceList);
   }
 
   private static byte[] makeLambda(String invokedName,
                                    MethodType invokedType,
                                    MethodType methodType,
-                                   MethodHandle methodImplementation)
+                                   MethodHandle methodImplementation,
+                                   Class[] interfaces)
   {
     String className;
     { int number;
@@ -220,8 +226,13 @@ public class LambdaMetafactory {
 
     List<PoolEntry> pool = new ArrayList();
 
-    int interfaceIndex = ConstantPool.addClass
-      (pool, invokedType.returnType().getName().replace('.', '/'));
+    int[] interfaceIndexes = new int[interfaces.length + 1];
+    interfaceIndexes[0] = ConstantPool.addClass(pool, invokedType.returnType().getName().replace('.', '/'));
+    for (int i = 0; i < interfaces.length; i++) {
+      String name = interfaces[i].getName().replace('.', '/');
+      System.out.println("interface name " + name);
+      interfaceIndexes[i + 1] = ConstantPool.addClass(pool, name);
+    }
 
     List<FieldData> fieldTable = new ArrayList();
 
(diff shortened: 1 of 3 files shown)


classpath/java/lang/invoke/LambdaMetafactory.java
```diff
@@ -220,8 +226,13 @@ private static String constructorSpec(MethodType type) {
 
     List<PoolEntry> pool = new ArrayList();
 
-    int interfaceIndex = ConstantPool.addClass
-      (pool, invokedType.returnType().getName().replace('.', '/'));
+    int[] interfaceIndexes = new int[interfaces.length + 1];
+    interfaceIndexes[0] = ConstantPool.addClass(pool, invokedType.returnType().getName().replace('.', '/'));
+    for (int i = 0; i < interfaces.length; i++) {
+      String name = interfaces[i].getName().replace('.', '/');
+      System.out.println("interface name " + name);
```
joshuawarner32: This looks like some println debugging... (should it be removed?)

classpath/java/lang/invoke/LambdaMetafactory.java
```diff
@@ -290,20 +319,48 @@ public static CallSite metafactory(MethodHandles.Lookup caller,
...
+    //    MethodType... bridges      // IF flags has BRIDGES set
+    //  )
+
+    MethodType methodType = (MethodType) args[0];
+    MethodHandle methodImplementation = (MethodHandle) args[1];
+
+    int flags = (Integer) args[3];
+    boolean serializable = (flags & FLAG_SERIALIZABLE) != 0;
+    if ((flags & FLAG_MARKERS) != 0)
+      throw new UnsupportedOperationException("FLAG_MARKERS");
```
joshuawarner32: Do you have an idea of what this is used for and what might be needed to implement it?  If so, it might be handy to leave a comment to that effect.

classpath/java/lang/invoke/LambdaMetafactory.java
```diff
@@ -290,20 +319,48 @@ public static CallSite metafactory(MethodHandles.Lookup caller,
...
+    if ((flags & FLAG_BRIDGES) != 0) {
+      int bridgeCount = (Integer) args[4];
+      if (bridgeCount > 0)
+        throw new UnsupportedOperationException("FLAG_BRIDGES");
```
joshuawarner32: Ditto.

src/machine.cpp
```diff
@@ -6034,11 +6066,14 @@ GcJclass* getDeclaringClass(Thread* t, GcClass* c)
 
 GcCallSite* resolveDynamic(Thread* t, GcInvocation* invocation)
 {
+  // This support for invokedynamic is limited to support for the LambdaMetafactory, as the method dispatch logic
+  // assumes the first three args are always as expected by that class. It doesn't do full indy support.
```
joshuawarner32: Nitpick: I assume by 'indy' you mean invokedynamic - but that might not be entirely clear to others, especially given the lack of precedent in the codebase.  Could you expand that name?

src/machine.cpp
```diff
@@ -6066,10 +6099,16 @@ GcCallSite* resolveDynamic(Thread* t, GcInvocation* invocation)
       = t->m->classpath->makeString(t, nameBytes, 0, nameBytes->length() - 1);
   PROTECT(t, name);
 
+  // This is the type of the linked-to method (e.g. lambda).
   GcMethodType* type = makeMethodType(
       t, c->loader(), invocation->template_()->spec(), 0, 0, 0);
   PROTECT(t, type);
 
+  // There are always 3 arguments to an indy bootstrap method on top of the extras in the bootstrap array (if any),
```
joshuawarner32: Ditto (indy).

src/machine.cpp
```diff
@@ -6157,6 +6201,68 @@ GcCallSite* resolveDynamic(Thread* t, GcInvocation* invocation)
     ++i;
   }
 
+  object varArgsArray_ = makeObjectArray(t, numVarArgs);
+  PROTECT(t, varArgsArray_);
+
+  if (numVarArgs > 0) {
+    assertT(t, i < bootstrapArray->length());
+    GcArray *varArgsArray = reinterpret_cast<GcArray*>(varArgsArray_);
```
joshuawarner32: I think this needs a `PROTECT(t, varArgsArray)` after it.

src/machine.cpp
mikehearn: Actually I realised that the comment is wrong and removed it - the other commit adding it back is some merge issue. The support is correct and general (well, not fully complete, but not entirely lambda specific). I'll get it fixed.

src/machine.cpp
joshuawarner32: Awesome, I'd love to read that!

Yeah, PROTECT is definitely fickle.  It's also one of those things whose misuse to the damnedest bugs.

They key insight is that we don't trigger a GC unless we're actually trying to allocate something -  which means there's no way for a GC to happen between creating that variable and calling PROTECT.  It's been a while since I worked in that code, but IIRC, there's something preventing us from considering a thread running VM code as being at a "safe-point", so other threads also can't trigger a GC.

In general, having an extra PROTECT call is a tiny performance problem, but missing one leads to things like memory corruption when the object gets moved but the pointer isn't updated.  So, it's better to err on the side of caution.

There's a PROTECT call in the generated `makeWhatever` constructors, so you can allocate one object and pass that to the constructor of another object, but you can't pass multiple unPROTECTed objects to a constructor, since constructing the second one of those might trigger a GC.

As a side-note, I've been a huge fan of Rust (the language) for a while - and one of the features that's under discussion right now is [enabling GC integration in the compiler](http://manishearth.github.io/blog/2016/08/18/gc-support-in-rust-api-design/), which would make it dramatically easier to write GC'd language runtimes in rust.  Not that there's any plan to translate avian into rust.  Honest.

mikehearn: Thanks. That clears things up. In this case though, varArgsArray is just a cast and varArgsArray_ is already protected. So I think we're good here?

mikehearn:
I've added comments to address the review. I think the PROTECT is unnecessary as the same pointer was PROTECTed just a few lines above.

One question that has come to mind is why Avian uses its own j.l.i package even in openjdk-src builds. I don't see anything in the OpenJDK implementation of the LambdaMetaFactory that's obviously hard to support, but it's presumably done this way for a reason. Implementing support for bridge methods would be kind of annoying I guess, and it's unclear why it'd be necessary for OpenJDK class lib users.

WRT rust: yes, Rust is ideal for this sort of work. That said, meta-circular VMs are also interesting. Cliff Click (architect of HotSpot's C2 compiler) has said if he were to write a JVM again he'd write it in Java.

I'm using Avian because it's so easy to statically link it into a single binary, which is useful for my current project because I need a JVM that can run without an operating system.

src/machine.cpp
joshuawarner32: If avian's GC were a non-compacting GC, all it would need to know about is the set of live objects so it can figure out what's not live and free that.  If that were the case, you would be right.

However, avian's GC is actually a compacting GC (at least in the first generation), meaning that it not only needs to know about all the live objects, but also all the pointers thereto so that it can update them to the new location.

Now, it's possible that the compiler would actually combine those two `GcArray*`/`GcObject*` variables into one (since it knows they're the same anyway) - but I think that's very unlikely since it changes the semantics of the program if the pointer-to-pointer in the PROTECT were ever actually stored to somewhere in that function.  Without that optimization, only one variable would change, but with it, both would change. (Assuming we leave it as is.)

If there's a garbage collection while executing that function, the object pointed to by `varArgsArray` might be moved - and if `varArgsArray` is not PROTECTed, it won't be updated to the new location, leaving us with a pointer to who-knows-where.

------

Also, I just noticed you're doing a `reinterpret_cast` here, which works fine in the happy path.  If you were casting an unknown object rather than an object you created, I would be more particular about using the checked (at least in debug builds) `cast` function rather than the `reinterpret_cast` function.

I think the right solution is to `cast` / `reinterpret_cast` the original return value from `makeObjectArray`, eliminating the need for the second variable.  I would point out, however, that all the other cases where we're calling `makeObjectArray` and then initializing it, we're actually just using the unsafe/untyped `setField` function to do that, without any casts.  I don't remember exactly why I left those that way I left those that way, but it *might* be significant.

That said, if this works in practice and we have tests covering it (which it looks like we might not...), I'm standing on thin ice.

Would it be difficult to add a test or two that hits this code path?  Or should the existing tests hit this?

dicej:
@mikehearn, I originally tried to use OpenJDK's java.util.invoke, but the interface between the Java code and the VM for that package is undocumented and has changed significantly in both major and minor releases of OpenJDK.  Rather than try to reverse engineer it and keep up with all the changes, it was much easier just to write our own minimal implementation.

Of course, if anyone were to volunteer to port Avian to the OpenJDK implementation and maintain that port, that would be awesome.  It's a big job, though.

dicej:
One more comment about the `PROTECT` macro: a pointer needs to be `PROTECT`ed if there is a chance that the VM thread will enter the "idle" state between the point where the pointer is set to a value and the point where that value is used.  As Josh mentioned, the GC is in Avian is not preemptive, so we don't need to worry about being interrupted at an arbitrary point in code, but entering the idle state means potentially yielding control to a GC cycle initiated by either the current thread or a different thread.

The idle state can be reached any time a thread tries to allocate memory from the Java heap, any time it blocks (e.g. to acquire a mutex or wait for an event), any time it calls out to code outside the VM (e.g. JNI methods), and at each backward branch in Java code.  And, of course, any time it calls a function which might directly or indirectly do one of those things.

mikehearn:
Thanks. I'll record that in the hacking guide. Too bad about the jli implementation, it's a big chunk of code. How does the current pull req look?

dicej:
@mikehearn: We still need to either add `PROTECT(varArgsArray);` or (preferably) eliminate that variable and instead do `GcArray* varArgsArray_ = cast<GcArray>(makeObjectArray(t, numVarArgs));`

joshuawarner32:
I would also like to make sure we have some test coverage before merging.

mikehearn:
Ah, right, PROTECT means "update this pointer if the object moves", not "pin this object so it doesn't move". OK, thanks.

Yes I'll add some tests when I get back to this PR. It might be a few days though, as I'm trying to get my project to a working state ASAP.

mikehearn:
@dicej Is this PR still relevant after #511? It looks like you took a different approach and didn't implement full varargs support in resolveDynamic but did a lot more work on the compiler instead.

Sorry for taking so long to get back to it, by the way. I just started returning to this work and noticed you had done some of it after this PR was opened.

dicej:
@mikehearn, yes I think this is still relevant.  I only did the minimum necessary to fix the lambda issue Luc reported on the Google group (and only realized I was partially duplicating your work after I was already deep into it).  This PR does a lot more.

As Josh mentioned, our main concern is to have some test coverage for the extra features you've implemented, both to make sure they work and so we don't accidentally break them in the future.

mikehearn:
Yes, OK. This PR also includes a reflection fix. Perhaps that should be split out. I've asked one of my team to pick this up so it may be another week or two before we have updated patches for you.

dicej:
I'm going to close this one given that it seems to be superseded by more recent PRs.  Feel free to reopen it if you disagree.

mikehearn:
Not at all - exfalso works for me and has been breaking up this PR and rebasing it. Thanks.

PR #509, opened 2016-10-23
Create a version api file

AdrianVovk:
Use this for automatic build tools to check for updates.

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,5 +1,5 @@
 {
-  "version":"1.2.0"
-  "downloadUrl":"https://github.com/ReadyTalk/avian/archive/v1.2.0.zip"
+  "version":"1.2.0",
+  "downloadUrl":"https://github.com/ReadyTalk/avian/archive/v1.2.0.zip",
   "downloadTar":"https://github.com/ReadyTalk/avian/archive/v1.2.0.tar.gz"
 }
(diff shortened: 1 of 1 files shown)


joshuawarner32:
If I understand you correctly, you should already be able to do this by monitoring for new tags in the repo.  I'm hesitant to do anything that makes the release process more complicated.

Can you explain a little more about the problem you're trying to solve on your end?  Are you just looking for (automated) notifications of updates?  IMO, that would be solved much better by some method already supported in an existing ecosystem (e.g. a package manager).

Also be aware that avian is deep in maintenance mode right now; major updates will be few and far between.  At this point, you may just be better off tracking master.

AdrianVovk:
@joshuawarner32 I am building an IDE that packages in Avian and Gradle together (separately) and uses some custom build scripts to make them work together. I see what you are saying though. I could just track the tags and just convert it into a format I can use in my existing code. I will be closing this.

Though, some sort of simplified version json file would be nice, for any future developers. Something like services.gradle.org would be nice for integrators, but I see what you are saying.

joshuawarner32:
I agree, such a JSON file would be nice - but I think it's only really valuable if it becomes (or is already) part of a systematic pattern shared across a bunch of projects.  I'm not opposed to a change like this; I'd just like to see that it's either pluggable into an existing packaging system or there's a well-reasoned argument that it reasonably stands to become a convention.  One-off systems and protocols are no fun to work with, on either the producer or consumer end.

Lastly, don't let anything I say dampen your enthusiasm. :)

PR #511, opened 2016-12-05
Lambda fixes

dicej:
For lambdas that implement java.io.Serializable, the compiler emits
calls to LambdaMetaFactory.altMetafactory, not
LambdaMetaFactory.metafactory, so I've provided a stub implementation
that currently ignores the extra parameters it receives.

This also fixes a bug in compiling lambda glue code for lambdas that
take longs and/or doubles.

### DIFF
--- a/classpath/java/lang/invoke/SerializedLambda.java
+++ b/classpath/java/lang/invoke/SerializedLambda.java
@@ -1,0 +1,48 @@
+/* Copyright (c) 2008-2016, Avian Contributors
+
+   Permission to use, copy, modify, and/or distribute this software
+   for any purpose with or without fee is hereby granted, provided
+   that the above copyright notice and this permission notice appear
+   in all copies.
+
+   There is NO WARRANTY for this software.  See license.txt for
+   details. */
+
+package java.lang.invoke;
+
+public class SerializedLambda implements java.io.Serializable {
+  public int getImplMethodKind() {
+    // todo
+    return 0;
+  }
+
+  public String getImplClass() {
+    // todo
+    return null;
+  }
+
+  public String getImplMethodName() {
+    // todo
+    return null;
+  }
+
+  public String getImplMethodSignature() {
+    // todo
+    return null;
+  }
+
+  public String getFunctionalInterfaceClass() {
+    // todo
+    return null;
+  }
+
+  public String getFunctionalInterfaceMethodName() {
+    // todo
+    return null;
+  }
+
+  public String getFunctionalInterfaceMethodSignature() {
+    // todo
+    return null;
+  }
+}
--- a/src/compile.cpp
+++ b/src/compile.cpp
@@ -4040,15 +4040,26 @@ bool isLambda(Thread* t,
   return vm::strcmp(reinterpret_cast<const int8_t*>(
                         "java/lang/invoke/LambdaMetafactory"),
                     bootstrap->class_()->name()->body().begin()) == 0
-         and vm::strcmp(reinterpret_cast<const int8_t*>("metafactory"),
-                        bootstrap->name()->body().begin()) == 0
-         and vm::strcmp(
-                 reinterpret_cast<const int8_t*>(
-                     "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/"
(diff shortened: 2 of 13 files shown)


dicej:
And it looks like I broke the non-openjdk builds.  Working on it.

dicej:
Hmm, and I just noticed https://github.com/ReadyTalk/avian/pull/505 covers the same ground and then some.  I should take a closer look at that.

dicej:
The bootimage build is still broken, and fixing it isn't as simple as I thought.  I'll work on it tomorrow.

dicej:
The build is fixed now.

src/compile.cpp
```diff
@@ -5133,120 +5144,147 @@ void compile(MyThread* t,
...
+
+              jclass lmfClass = e->vtable->FindClass(
+                  e, "java/lang/invoke/LambdaMetafactory");
+              jmethodID makeLambda = e->vtable->GetStaticMethodID(
+                  e,
+                  lmfClass,
+                  "makeLambda",
+                  "(Ljava/lang/String;Ljava/lang/String;Ljava/lang/"
+                  "String;Ljava/"
+                  "lang/String;Ljava/lang/String;Ljava/lang/String;I)[B");
```
joshuawarner32: (What I assume is clang-format) didn't do a great job here.  I'm slowly learning all the ways in which auto-formatting is not a panacea...

classpath/java/lang/invoke/SerializedLambda.java
```diff
@@ -0,0 +1,48 @@
...
+   for any purpose with or without fee is hereby granted, provided
+   that the above copyright notice and this permission notice appear
+   in all copies.
+
+   There is NO WARRANTY for this software.  See license.txt for
+   details. */
+
+package java.lang.invoke;
+
+public class SerializedLambda implements java.io.Serializable {
```
joshuawarner32: Not really a decision we can make as downstream spec implementors, but it seems to me that serializing lambdas is just asking for some sort of security vulnerability.

test/InvokeDynamic.java
```diff
@@ -24,8 +42,19 @@ public static void main(String[] args) {
   }
 
   private void test() {
-    int c = 2;
-    Operation op = (a, b) -> ((a + b) * c) - foo;
-    expect(op.operate(2, 3) == ((2 + 3) * 2) - foo);
+    { int c = 2;
+      Operation op = (a, b) -> ((a + b) * c) - foo;
```
joshuawarner32: Will this test get AOT-compiled at all?  I thought it was just the classpath.  If not, this may not yield good coverage.

dicej:
On Dec 5, 2016 6:32 PM, "Joshua Warner" <notifications@github.com> wrote:

*@joshuawarner32* approved this pull request.

My only concern is if the test code gives any coverage of the AOT
compilation. Can you confirm? (Or otherwise argue that it's not necessary /
worth-it)


Yes, that's what bootimage-test does in ci.sh.  It compiles the tests into
a bootimage and runs them that way.  Hence the broken build prior to my
compile.cpp work.

------------------------------

In test/InvokeDynamic.java
<https://github.com/ReadyTalk/avian/pull/511#pullrequestreview-11506418>:

> @@ -24,8 +42,19 @@ public static void main(String[] args) {
   }

   private void test() {
-    int c = 2;
-    Operation op = (a, b) -> ((a + b) * c) - foo;
-    expect(op.operate(2, 3) == ((2 + 3) * 2) - foo);
+    { int c = 2;
+      Operation op = (a, b) -> ((a + b) * c) - foo;

Will this test get AOT-compiled at all? I thought it was just the
classpath. If not, this may not yield good coverage.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ReadyTalk/avian/pull/511#pullrequestreview-11506418>,
or mute the thread

joshuawarner32:
*Ah*, yep.  I forgot about the `bootimage-test=true` step.

joshuawarner32:
(LGTM; merge at your leisure)

src/compile.cpp
dicej: It's partly my fault, since there are so many levels of indentation and we're limiting clang-format to 80 columns.  It wouldn't be quite as bad if I broke this into several methods.

PR #514, opened 2016-12-16
Added basic support for the Runtime.maxFree() method.

siepkes:
Added basic support for the `Runtime.maxFree()` method.  Some applications (BouncyCastle for example) expect to be able to call this method. It is implemented basically the same as `totalMemory()` since there isn't much of a difference for Avian.

Give me a ping if any changes are needed.

### DIFF
--- a/classpath/java/lang/Runtime.java
+++ b/classpath/java/lang/Runtime.java
@@ -129,6 +129,8 @@ public class Runtime {
   public native long freeMemory();
 
   public native long totalMemory();
+  
+  public native long maxMemory();
 
   private static class MyProcess extends Process {
     private long pid;
--- a/src/builtin.cpp
+++ b/src/builtin.cpp
@@ -372,6 +372,12 @@ extern "C" AVIAN_EXPORT int64_t JNICALL
 }
 
 extern "C" AVIAN_EXPORT int64_t JNICALL
+    Avian_java_lang_Runtime_maxMemory(Thread* t, object, uintptr_t*)
+{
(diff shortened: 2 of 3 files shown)


joshuawarner32:
Not a big issue, but you have the correct `maxMemory` name in the code, and you're calling it `maxFree` in the commit and branch name.  That threw me off for a moment. :)

Reading through the documentation, I think `maxMemory` is doing what it should here, and `totalMemory` is actually supposed to report how much of that max memory is actually allocated from the OS right now.  It will tend to approach `maxMemory` over time.  In other words, there should be a new method on `Heap` (say, `allocated`) that returns `c.gen1.capacity_ + c.gen2.capacity`.  

I don't think that needs to be fixed here.  Just a thought.

siepkes:
@joshuawarner32 Can imagine that might cause some confusion, sorry about that.

Yeah the actual meaning of those methods is open for debate ;-). The doc says: "the maximum amount of memory that the virtual machine will attempt to use, measured in bytes". That sounds an awful lot like the max heap size. Then again it doesn't say it is limited to the heap. Should off heap memory (like permgen for pre Java 8) count towards this max?

Like you said I don't think that needs to be fixed here. For me the primary purpose of this pull request was to increase compatibility, which this merge does. But I agree with you that looking in to the behavior of the whole `freeMemory`, `totalMemory` and `maxMemory` set of methods is a good idea.

joshuawarner32:
Thanks @siepkes!

PR #520, opened 2017-02-06
make OpenJDK build compatible with ojdkbuild artifacts

dicej:
One way to avoid the intense pain of building OpenJDK on Windows is to
let somebody else do it, e.g. https://github.com/ojdkbuild/ojdkbuild.
However, that project customizes the builds slightly, which means we
need to make an effort to stay compatible.  This one-liner does that.

### DIFF
--- a/src/classpath-openjdk.cpp
+++ b/src/classpath-openjdk.cpp
@@ -669,6 +669,10 @@ class MyClasspath : public Classpath {
 #  ifdef PLATFORM_WINDOWS
     expect(t, loadLibrary(t, libraryPath, "msvcr100", true, true));
 #  endif
+    // necessary for using OpenJDK builds from
+    // https://github.com/ojdkbuild/ojdkbuild:
+    loadLibrary(t, libraryPath, "ojdkbuild_zlib", true, true, false);
+
     expect(t, loadLibrary(t, libraryPath, "verify", true, true));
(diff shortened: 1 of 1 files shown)


src/classpath-openjdk.cpp
```diff
@@ -669,6 +669,10 @@ class MyClasspath : public Classpath {
 #  ifdef PLATFORM_WINDOWS
     expect(t, loadLibrary(t, libraryPath, "msvcr100", true, true));
 #  endif
+    // necessary for using OpenJDK builds from
+    // https://github.com/ojdkbuild/ojdkbuild:
+    loadLibrary(t, libraryPath, "ojdkbuild_zlib", true, true);
```
joshuawarner32: Serious edge case: Could this cause problems if the library is there, but we're not supposed to link it?  I.e. Perhaps the user has multiple installs of openjdk, one from ojdkbuild and one not.

dicej:
Yeah, there is the potential danger that there is a ojdkbuild_zlib.dll in the user's library path at runtime that shouldn't be used.  One possible improvement would be to explicitly check for the presence of that file at build time, and only try to load it at runtime if it was there at build time, but I don't think it's worth the effort.  I don't think anyone's seriously using openjdk=... without openjdk-src=... anyway.

PR #522, opened 2017-02-07
Add optional debugging of method calls

exFalso:


src/interpret.cpp
```diff
@@ -574,10 +574,11 @@ unsigned invokeNativeSlow(Thread* t, GcMethod* method, void* function)
   uint64_t result;
 
   if (DebugRun) {
+    signed char *cname = method->class_() && method->class_()->name() ? method->class_()->name()->body().begin() : (signed char*) "?";
```
dicej: We should be able to declare `cname` as a `const char*` and avoid the cast here and elsewhere.  Also, we prefer to avoid C-style casts, since they're too general (i.e. they don't communicate the intent of the programmer as well as the more specific C++ casts).  Better to use static_cast, const_cast, or reinterpret_cast depending on the goal.

src/interpret.cpp
```diff
@@ -2989,15 +2991,21 @@ invoke : {
   if (method->flags() & ACC_NATIVE) {
     invokeNative(t, method);
   } else {
+    if (DebugCalls && method) {
```
dicej: Doesn't seem necessary to check that `method` is non-null here, since we dereferenced it above.

src/interpret.cpp
dicej: I guess since `method->class_()->name()->body().begin()` returns an `int8_t*`, you'll still need a reinterpret_cast in there, but at least you should be able to avoid a const_cast by adding `const` to the declared type.

PR #525, opened 2017-02-15
Added functionality to HttpURLConnection for obtaining content length

MaartenR:
In order to be able to keep track of the progress of a file download we needed a way to obtain the size of this file beforehand. Hence this pull request. A map with http headers was already being created, but there was no public api available for retrieving the values. Therefore i added the methods we needed.


MaartenR:
Sorry, took a while, but http headers should now be case insensitive.

PR #527, opened 2017-03-13
Added numberOfLeadingZeros method to Integer class

MaartenR:
Another small utility method that was needed for our project. This time the numberOfLeadingZeros method of the integer class. The source of the algorithm is included in a comment.

### DIFF
--- a/classpath/java/lang/Integer.java
+++ b/classpath/java/lang/Integer.java
@@ -139,4 +139,14 @@ public final class Integer extends Number implements Comparable<Integer> {
     }
     return new Integer(parseInt(string, 10));
   }
+
+  public static int numberOfLeadingZeros(int i) {
+    // See nlz5 at http://www.hackersdelight.org/hdcodetxt/nlz.c.txt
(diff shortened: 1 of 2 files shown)


classpath/java/lang/Integer.java
```diff
@@ -139,4 +139,14 @@ public static Integer decode(String string) {
     }
     return new Integer(parseInt(string, 10));
   }
+
+  public static int numberOfLeadingZeros(int i) {
```
dicej: Are you sure this is correct for a signed integer?  I would assume that it needs an unsigned shift right (>>>) at least.

dicej: It would be good to add a few test cases to the test suite to make sure this behaves correctly for e.g. negative numbers, very large positive numbers, etc.

MaartenR: I don't know about the right shift, but it seems that it shouldn't be an unsigned one. I agree that adding test cases would be a good idea, so i added some.

PR #535, opened 2017-05-30
Fixed issue with toArray on value/keySet in ConcurrentHashMap

### LINKED ISSUE #381: ConcurrentHashMap.<someCollection>.toArray() can throw ArrayIndexOutOfBoundsexception
I noticed this while looking for another issue.  If you do ConcurrentHashMap.entrySet/values/keySet, and then try to call .toArray (weather an array is passed in or not).  If the map structure grows after the .size() check in Data.toArray, the iterator will end up getting an extra element than the array that was allocated for it.  Thus the out of bounds exception will be thrown when we try to store an element in the array we have not allocated for.

lwahlmeier:
fixes #381

### DIFF
--- a/classpath/avian/Data.java
+++ b/classpath/avian/Data.java
@@ -15,7 +15,6 @@ import java.util.Map.Entry;
 import java.util.AbstractSet;
 import java.util.Collection;
 import java.util.Iterator;
-import java.util.Collections;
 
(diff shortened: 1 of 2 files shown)


dicej:
Closing this due to the Travis test failure.

PR #530, opened 2017-05-14
More compatibility with Java 8 lambdas

mikehearn:

PR #539, opened 2017-07-03
Add support for "invokevirtual" and "invokeinterface".

chrisr3:
Trying to use a function reference on Avian (compiled against OpenJDK):
```java
List<String> list = asList(args);
int totalChars = list.stream().mapToInt(String::length).sum();
```

resulted in this error:

todo: implement per [http://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.5](http://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.5)

Following the link revealed that Avian lacked support for `invokevirtual`, which is now trivially(?) fixed. For an encore, I then implemented `invokeinterface` as well.

### DIFF
--- a/classpath/avian/ConstantPool.java
+++ b/classpath/avian/ConstantPool.java
@@ -26,6 +26,7 @@ public class ConstantPool {
   private static final int CONSTANT_NameAndType = 12;
   private static final int CONSTANT_Fieldref = 9;
   private static final int CONSTANT_Methodref = 10;
+  private static final int CONSTANT_InterfaceMethodref = 11;
   
   public static int add(List<PoolEntry> pool, PoolEntry e) {
     int i = 0;
@@ -83,6 +84,16 @@ public class ConstantPool {
     return add(pool, new MethodRefPoolEntry
                (addClass(pool, className),
                 addNameAndType(pool, name, spec)));
+  }
+
+  public static int addInterfaceMethodRef(List<PoolEntry> pool,
+                                          String interfaceName,
+                                          String name,
+                                          String spec)
+  {
+    return add(pool, new InterfaceMethodRefPoolEntry
+               (addClass(pool, interfaceName),
+               addNameAndType(pool, name, spec)));
   }
 
   public interface PoolEntry {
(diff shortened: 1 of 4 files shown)


chrisr3:
This looks like an enhancement to #530. The required test-case is as above.

dicej:
Thanks for doing this, Chris.

Would you mind adding tests for both the invokevirtual and invokeinterface cases to your PR?  We already have a number of lambda tests in test/InvokeDynamic.java, so that would be a good place to add them.

chrisr3:
I have found this test case for `invokeinterface`:
```java
BiFunction<CharSequence, Integer, Character> f = CharSequence::charAt;
String data = "0123456789";
for (int i = 0; i < data.length(); ++i) {
    System.out.println(f.apply(data, i));
}
```
Unfortunately, I've now realised that my implementation here is broken. I am using [this](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.invokeinterface) document as a reference for a correct implementation, but am wondering how to derive the `count` byte that it mentions. My initial thought was that it would be `1 + method.parameterFootprint`. However, Avian doesn't seem to account for each `long` or `double` parameter contributing **2** to the final footprint value, as described in the [Notes](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.invokeinterface.notes). Am I interpreting `parameterFootprint` correctly here, please?

It's entirely possible that this doesn't matter, because `long` and `double` would almost certainly need to be boxed for these lambdas to work, but I would like to be sure that I've understood this.

chrisr3:
All of the test cases involving function references that use native types are currently failing. However, all of these cases execute correctly with other JVMs.

chrisr3:
The problem I'm seeing here with native types is more subtle than I had originally suspected. E.g.
```java
public class TestInvokeInterfaceLongBad {

  private interface BiFunction<T, U, R> {
    R apply(T t, U u);
  }

  private interface GetLong {
    long get(long l);
  }

  private static class LongHolder implements GetLong {
    @Override
    public long get(long l) {
        return l;
    }
  }

      BiFunction<GetLong, Long, Long> f = GetLong::get;
      System.out.println("Long: " + f.apply(new LongHolder(), 20L));
  }
}
```
This fails because Avian cannot find interface `TestInvokeInterfaceLongBad$GetLong` on its `Lambda-0` class. However, if I rewrite the `GetLong` interface to use `Long` instead of native `long` then Avian searches for `TestInvokeInterfaceLongGood$GetLong` on `TestInvokeInterfaceLongGood$LongHolder` instead. And this works.

dicej:
Hi Chris,

Thanks for working on this, and sorry I haven't been very responsive.  I'll investigate the issues you found this weekend when I have time.

And yes, `parameterFootprint` should account for double and long parameters taking two slots.

dicej:
This fixes all the test cases you added: https://github.com/ReadyTalk/avian/commit/ecafdf40baf2f3cde4835620e9af4380b1081fe7

Thanks for finding those cases and taking the first steps towards fixing them.

PR #541, opened 2017-07-27
Fix crash when resolving system classes.

chrisr3:
I uncovered this crash when trying to write a classloader that extends `avian.SystemClassLoader`. The problem is that `loader->map()` can return `NULL`, but the `resolveSystemClass()` function doesn't check for this. However, this check _is_ built into the `findLoadedClass()` function, so use that instead.

Having said this, I am wondering if `avian.SystemClassLoader` is _supposed_ to be extended in the first place. Consider this test case, which currently fails because the class returned by `findClass()` doesn't have `myLoader` as its classloader, despite the class having been loaded by it:
```java
import java.io.IOException;
import java.io.File;
import java.io.FileInputStream;

public class CustomClassLoaders {
  private static void expect(boolean v) {
    if (! v) throw new RuntimeException();
  }

  private static void testClassLoader() throws Exception {
    MyClassLoader myLoader = new MyClassLoader();
    Class<?> myClass = myLoader.findClass("CustomClassLoaders$Hello");
    expect(myClass.getClassLoader() == myLoader);
  }

  public static void main(String[] args) throws Exception {
    testClassLoader();
  }

  public static class Hello {
      System.out.println("hello, world!");
    }
  }

  private static class MyClassLoader extends avian.SystemClassLoader {
    @Override
    public Class<?> findClass(String name) throws ClassNotFoundException {
      return super.findClass(name);
    }
  }
}
```
The OpenJDK equivalent of `MyClassLoader` would look like this:
```java
  private static class MyClassLoader extends java.net.URLClassLoader {
    MyClassLoader(ClassLoader parent) {
        super(((URLClassLoader)parent).getURLs(), parent);
    }

    @Override
    public Class<?> findClass(String name) throws ClassNotFoundException {
      return super.findClass(name);
    }
  }
```
And this _does_ associate class `Hello` with my custom classloader, so Avian's classloading semantics look different to me.

### DIFF
--- a/src/machine.cpp
+++ b/src/machine.cpp
@@ -4881,13 +4881,7 @@ GcClass* resolveSystemClass(Thread* t,
 
   ACQUIRE(t, t->m->classLock);
 
-  GcClass* class_ = cast<GcClass>(t,
-                                  hashMapFind(t,
-                                              cast<GcHashMap>(t, loader->map()),
-                                              spec,
-                                              byteArrayHash,
-                                              byteArrayEqual));
-
+  GcClass* class_ = findLoadedClass(t, loader, spec);
   if (class_ == 0) {
     PROTECT(t, class_);
 


chrisr3:
I have hacked together a "Nuclear Evil" native method to alter a class's classloader immediately after it has been resolved:
```c++
extern "C" AVIAN_EXPORT void JNICALL
    Avian_avian_Classes_nuclearEvil(Thread* t, object, uintptr_t* arguments)
{
  GcJclass* classType = cast<GcJclass>(t, reinterpret_cast<object>(arguments[0]));
  GcClassLoader* classLoader = cast<GcClassLoader>(t, reinterpret_cast<object>(arguments[1]));
  classType->vmClass()->setLoader(t, classLoader);
  classType->setClassLoader(t, classLoader);
}
```
Avian does behave as expected after this, using my new classloader to link the class. However, it would obviously be nicer if Avian set the classloader automatically.

dicej:
Hi Chris,

Sorry about the delayed response.  As you found, SystemClassLoader was never meant to be used directly or extended by application code, and the same goes for almost all the classes under the `avian` package.  Those classes exist to bridge the abstraction gap between the Avian VM and the standard Java classes.

Did you try extending java.net.URLClassLoader as you would have with OpenJDK?  If you did try, and it didn't work, maybe there's something we can fix there so you don't have to reference a non-standard class like SystemClassLoader in your application code.

chrisr3:
Hi, thanks for replying.
My use-case is trying to filter which classes can be loaded _from the boot classloader itself_, which I have been unable to do by writing Java class-loaders. I have ultimately resorted to implementing my filter inside here instead:
```C++
GcClass* resolveSystemClass(Thread* t,
                            GcClassLoader* loader,
                            GcByteArray* spec,
                            bool throw_,
                            Gc::Type throwType);
```
This seems to have the semantics that I want, even if I did also have to modify `avian.SystemClassLoader` directly itself.

dicej:
I see.  So how would you have done that on OpenJDK?  I would guess that there's no standard way to do it, in which case it makes sense that you'd have to use a non-standard, VM-specific approach.

chrisr3:
Yes, I found a similar place to patch such a filter into HotSpot once I'd given up trying to inject a new system classloader by setting OpenJDK's `java.system.class.loader` property. IIRC, I had tried implementing it as:
```Java
class MyClassLoader extends URLClassLoader {
    MyClassLoader(ClassLoader parent) {
        super(((URLClassLoader)parent).getURLs(), parent);
    }

    ...
```
which is obviously incompatible with Avian anyway because `avian.SystemClassLoader` doesn't extends `URLClassLoader`.

PR #545, opened 2017-08-29
Support Thread.uncaughtExceptionHandler with OpenJDK

chrisr3:
I have noticed that Avian doesn't support `Thread.uncaughtExceptionHandler` properly when compiled with OpenJDK. ~~This PR _almost_ does what I need except that I have noticed that the test cases are failing because `Thread.join()` isn't waiting for the `uncaughtExceptionHandler` to be executed.~~

~~The test case passes when Avian is _not_ compiled against OpenJDK, e.g.~~

> $ make clean test

~~My test case also passes when executed by both Oracle's JVM and OpenJDK itself, and all JVMs (including Avian) execute `uncaughtExceptionHandler` in the child thread. It therefore seems reasonable for `Thread.join()` to wait until the handler has executed too.~~

I have refactored how this works so that `uncaughtExceptionHandler` is now invoked _before_ `ActiveFlag` is cleared. This allows `Thread.join()` to wait correctly.

### DIFF
--- a/src/classpath-openjdk.cpp
+++ b/src/classpath-openjdk.cpp
@@ -981,11 +981,16 @@ void uncaughtException(Thread *t, GcThrowable *e)
                                      "(Ljava/lang/Throwable;)V");
   if (dispatch != NULL) {
     THREAD_RESOURCE0(t, {
-      // We ignore any exceptions from the uncaught
-      // exception handler itself.
-      t->exception = NULL;
+      if (t->exception != NULL) {
+        // We ignore any exceptions from the uncaught
+        // exception handler itself.
+        t->exception = NULL;
 
-      disposeThread(t);
+        // The stack will be unwound when this resource is
+        // released, which means that uncaughtException()
+        // will not return. So repeat the thread clean-up here.
+        disposeThread(t);
+      }
     });
 
     t->m->processor->invoke(t, dispatch, t->javaThread, e);
--- a/src/classpath-openjdk.cpp
+++ b/src/classpath-openjdk.cpp
@@ -296,6 +296,10 @@ object makeJconstructor(Thread* t, GcMethod* vmMethod, int index = -1);
 
 object makeJfield(Thread* t, GcField* vmField, int index = -1);
 
+void uncaughtException(Thread *t, GcThrowable *e);
+
+void disposeThread(Thread *t);
+
 #ifdef AVIAN_OPENJDK_SRC
 void interceptFileOperations(Thread*, bool);
 #endif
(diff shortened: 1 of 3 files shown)


src/classpath-openjdk.cpp
```diff
@@ -975,6 +978,22 @@ class EmbeddedFile {
...
+                                     roots(t)->bootLoader(),
+                                     "java/lang/Thread",
+                                     "dispatchUncaughtException",
+                                     "(Ljava/lang/Throwable;)V");
+  if (dispatch != NULL) {
+    t->m->processor->invoke(t, dispatch, t->javaThread, e);
+
+    // We ignore any exceptions from the uncaught
+    // exception handler itself.
+    t->exception = NULL;
```
dicej: I don't think this is enough, since the invoke call above may unwind the stack past it.  Have you tested an exception handler that throws an exception?  I recommend adding such a test to ThreadExceptions.java.  We might need to add a THREAD_RESOURCE0 here to catch the exception and clear it.

dicej: e.g. something like this:

```c++
THREAD_RESOURCE0(t, {
  t->exception = NULL;
});

t->m->processor->invoke(t, dispatch, t->javaThread, e);
```

chrisr3:
Not sure what this build error is about, because the failing configuration doesn't have anything to do with `classpath-openjdk.cpp`.

src/classpath-openjdk.cpp
chrisr3: I see your point: when the `uncaughtExceptionHandler` throws an exception, the `t-checkpoint->unwind()` statement aborts the release of the previous `THREAD_RESOURCE0`. It looks like this might need a custom `Checkpoint` object. Or wrap the method inside a Java `try-catch` block.

dicej:
I'm going to assume the Travis build will succeed when this is merged into master.  If not, we'll fix it.

chrisr3:
Thanks for merging this, although the "exception code path" is rather hard to follow now. I'm not 100% certain how this call-stack unwinding mechanism works, but it looks we'd need to install a new "ignore exceptions" checkpoint before invoking `dispatchUncaughtException()` before my `uncaughtException()` function would be guaranteed to return in all cases. And unfortunately, such a checkpoint would seem to need to be implemented in assembler across multiple architectures!

Ideally, we'd like my new `THREAD_RESOURCE0` block to execute _and then return_ to allow the original `THREAD_RESOURCE0` block to _continue_ executing from where it left off.

dicej:
Yeah, it could definitely be refined.  I'll take another look when I have a chance.  If you look for ueses `vmRun` in the code, you can see examples of how to prevent a thread from unwinding past a given point.  You can think of vmRun as a way to catch exceptions and prevent the stack from unwinding further.

`THREAD_RESOURCE0`, on the other hand, is intended to be used to temporarily interrupt stack unwinding so some code can be run before the unwind is resumed.  I do think a direct call to `vmRun` might be more appropriate in this case, but I haven't had time yet to think it through.

PR #546, opened 2017-09-05
Fix "fallthrough" warnings with recent GCC.

chrisr3:
Recent GCC, e.g. in my case:

    gcc (GCC) 7.1.1 20170622 (Red Hat 7.1.1-3)

warns about `case` statements that "fall through" to the ones below. The solution is to comment them explicitly.


dicej:
Thanks, Chris!

PR #549, opened 2017-09-10
Changes return type to StringBuilder

seanhenry:
I get a method not found exception when using the `StringBuilder` because it returns an `Appendable` type rather than a `StringBuilder` type.


dicej:
Thanks Sean.  LGTM.

The Travis build failed do to a problem unrelated to your patch, so I'll address that seperately.

seanhenry:
Thanks!

PR #550, opened 2017-09-11
Refactor handling of uncaught exceptions with OpenJDK to use runRaw().

chrisr3:


src/classpath-openjdk.cpp
```diff
@@ -600,10 +598,24 @@ class MyClasspath : public Classpath {
         PROTECT(t, e);
```
dicej: We can remove this `PROTECT` line since the local variable `e` is not used after the call to runRaw.

src/classpath-openjdk.cpp
```diff
@@ -972,37 +984,19 @@ class EmbeddedFile {
...
-      if (t->exception != NULL) {
-        // The stack will be unwound when this resource is
-        // released, which means that uncaughtException()
-        // will not return. So repeat the thread clean-up here.
-        disposeThread(t);
-      }
-    });
-
-    t->m->processor->invoke(t, dispatch, t->javaThread, e);
+    GcThrowable* exception
```
dicej: We need to move this to before the call to resolveMethod and protect it:
```c++
GcThrowable* exception = cast<GcThrowable>(t, reinterpret_cast<object>(arguments[0]));
PROTECT(t, exception);
``` 
Otherwise, `resolveMethod` may trigger a GC and move the object `arguments[0]` points to without updating the pointer.

chrisr3:
OK, thanks. Both (hopefully) fixed now.

dicej:
Thanks, Chris!

PR #551, opened 2017-10-08
Support proper minimum version instead of fishing the latest version

teras:
The current scheme of iOS version has a few problems:

- It compiles as minimum version the latest version
- There is no way to define the minimum version
- The newest SDK supports only 64 bit mode, since 32 bit was dropped; so an older SDK version is required
- It tries to guess the latest version from a filename scheme, which is fragile and changes in any new SDK version. Actually this is not even necessary since there is a version-less location of the latest SDK.

Please consider this pull request as a base to discuss how these issues could be solved

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ certain flags described below, all of which are optional.
         use-clang={true,false} \
         openjdk=<openjdk installation directory> \
         openjdk-src=<openjdk source directory> \
-        android=<android source directory>
+        android=<android source directory> \
+        ios-version=<iOS minimum version>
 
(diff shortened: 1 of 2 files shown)


dicej:
Thanks for the PR.  This looks fine to me.  I haven't done any iOS development for a long time, so I don't have any specific opinions on the topic, and I agree that the old way was fragile.

Shall I merge this now, or is there more to discuss and/or change?

teras:
It works for me :) So if there is no objection, I'd suggest to merge it. If more issues arise, I'll be happy to have a look at it.

PR #571, opened 2020-08-10
Patch for multiple integer overflows (CVE-2020-17360) and silent return on negative length (CVE-2020-17361)

polivar3:
This patch will fix the following security issues in arrayCopy:
-Two integer overflows resulting in out-of-bounds read/write (CVE-2020-17360).
-A silent return when negative lengths are provided. This could result in data being lost during the copy, with varying consequences depending on the subsequent use of the destination buffer (CVE-2020-17361).

### DIFF
--- a/src/avian/classpath-common.h
+++ b/src/avian/classpath-common.h
@@ -104,7 +104,7 @@ void arrayCopy(Thread* t,
             throwNew(t, GcIndexOutOfBoundsException::Type);
           }
         } else if (LIKELY(length < 0)) {
-          throwNew(t, GcNegativeArraySizeException::Type, "%d", length);
+          throwNew(t, GcIndexOutOfBoundsException::Type, "%d", length);
           return;
         } else {
           return;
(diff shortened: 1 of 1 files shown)


src/avian/classpath-common.h
```diff
@@ -103,6 +103,9 @@ void arrayCopy(Thread* t,
           } else {
             throwNew(t, GcIndexOutOfBoundsException::Type);
           }
+        } else if (LIKELY(length < 0)) {
+          throwNew(t, GcNegativeArraySizeException::Type, "%d", length);
```
dicej, line 107: According to https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#arraycopy(java.lang.Object,%20int,%20java.lang.Object,%20int,%20int), this should throw an `IndexOutOfBoundsException` if the length is negative.

polivar3: Good catch! I've made the change to adhere to the Oracle docs.

dicej:
For the record, CI is failing because Travis is now using JDK 11, which Avian's build system has trouble with.  Not sure how much work would be required to fix that.  Not a blocker for this PR, anyway.

dicej:
Thanks, @polivar3!

polivar3:
Thanks to you @dicej

